Issue #37, opened 2019-03-29
Add --template flag to inject command


SimonBarendse:
When using a template from a file (which is the large majority of use-cases), you currently have to `cat` the file and pipe it to the inject command:

`cat env.tpl | secrethub inject`

Like with #35, not all environments work nicely with pipes. For example, in a docker image, [we prefer a single call to a binary over the shell form of the command](https://docs.docker.com/engine/reference/builder/#cmd).

Therefore, I propose to add a `--template-file` flag that can be used as an alternative input to `stdin`.

```
--template-file (string)

   Use the contents of a file as template. 
```

And the order in which the command reads the template from different inputs would change to:

1. `--template-file`
1. If input is piped, read from stdin
1. Error if no other input is given

SimonBarendse:
An `--in-file` flag is added in [v0.24.0](https://github.com/secrethub/secrethub-cli/releases/tag/v0.24.0).

PR #47, opened 2019-04-04
Add a Dockerfile

SimonBarendse:


Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+FROM build_base as build
+RUN apk add --update make
+COPY . .
+RUN make build
+
+FROM alpine
+COPY --from=build /build/secrethub /usr/bin/secrethub
+RUN apk add --no-cache ca-certificates && \
+    update-ca-certificates
+ENTRYPOINT ["secrethub"]
```
florisvdg: With `secrethub` as `ENTRYPOINT`, can you test how well extending it works? And do we want to support it that you can `docker exec` into it at runtime and run ad hoc commands in a shell?

SimonBarendse: We're working on making it possible to run all commands without having to execute in shell form. For example, see https://github.com/secrethub/secrethub-cli/issues/37 for adding a `--template` flag to the `inject` command so that that can be run in exec form.

Note that in the meantime, you can override the entrypoint.

---

As for the docker exec: Do you mean to have an initial command that does not exit, so that `docker exec` can be called at arbitrary moments on the same container?

That might be useful in some cases. However, I don't think that should be the default. The image can be used like that by overriding the entrypoint.

PR #48, opened 2019-04-04
Let the signup command print that the email address should be verified

SimonBarendse:


internals/secrethub/signup.go
```diff
@@ -164,5 +164,6 @@ func (cmd *SignUpCommand) Run() error {
 	}
 
 	fmt.Fprintln(cmd.io.Stdout(), "Signup complete! You're now on SecretHub.")
+	fmt.Fprintf(cmd.io.Stdout(), "We've send an email to %s Please verify your email address to continue.\n", cmd.email)
```
florisvdg: `Please verify your email address to continue, we've sent a verification mail to %s.\n"`

SimonBarendse: Yes, that's better :+1:

PR #52, opened 2019-04-08
 Release on Snapcraft 

SimonBarendse:
Release on [Snapcraft](https://snapcraft.io/).

See https://github.com/goreleaser/goreleaser/pull/1000 for adding the `bin_directory` option on the archive instead of setting it in the binary name.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -25,7 +25,7 @@ archive:
   format_overrides:
     - goos: windows
       format: zip
-  bin_directory: yes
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -25,6 +25,7 @@ archive:
   format_overrides:
     - goos: windows
       format: zip
+  bin_directory: yes
```
florisvdg: More consistent to use `true` here

PR #56, opened 2019-04-09
Goreleaser: use secrethub as the snap package name

SimonBarendse:
The snap name is also the name with which you call the command. This should be `secrethub`.


florisvdg:
It says that the app name must be the same as the binary name. But can the snap name can then be different? Looks like it, because you can specify multiple apps in the same snap.

PR #58, opened 2019-04-09
Add script to generate Windows installers (.msi)

florisvdg:


Makefile
```diff
@@ -26,3 +26,6 @@ build:
 
 install:
 	go install ${BUILD_FLAGS} ./cmd/secrethub
+
+msi:
```
mackenbach: Consider adding the `make msi` command to the goreleaser config as a pre or post build hook (not sure which would be preferred here, I think post). Then we can include the resulting `.msi` files in the release.

msi.sh
```diff
@@ -0,0 +1,11 @@
+NAME="SecretHub CLI"
+EXE=bin/secrethub.exe
+VERSION=$(git describe --tags $(git rev-list --tags --max-count=1))
+MANUFACTURER=SecretHub
+UPGRADE_CODE=7178CB18-B431-4A9D-B29E-D31E63A1CF77
```
mackenbach: Is this a static value?

msi.sh
```diff
@@ -0,0 +1,11 @@
...
+ICON="/dist/secrethub.png"
+
+curl -s -o dist/secrethub.png https://secrethub.io/img/secrethub-logo-rgb-shield-square.png
+mkdir -p dist/msi
+docker run -v $(pwd)/dist:/dist msi-packager /dist/windows_amd64 /dist/msi/secrethub_windows_amd64.msi -n $NAME -e $EXE -v $VERSION -m $MANUFACTURER -u $UPGRADE_CODE -i $ICON --local
+docker run -v $(pwd)/dist:/dist msi-packager /dist/windows_386 /dist/msi/secrethub_windows_386.msi -n $NAME -e $EXE -v $VERSION -m $MANUFACTURER -u $UPGRADE_CODE -i $ICON --local
```
mackenbach: Wow this looks nice! Love how simple and isolated it is with the docker build container.

Makefile
florisvdg: I tried that at first, but those hooks very limited. Can have another look later.

msi.sh
florisvdg: Yes this is an arbitrary GUID to identify our app, so Windows knows it's same app across potential name changes.

florisvdg:
It works, but it's not that pretty. It puts an icon on the desktop and it doesn't symlink the binary.

msi.sh
mackenbach: Ah ok, got it 👍 Consider adding a comment as this is some weird arcane MSI knowledge.

mackenbach:
Hmm, ok. Let's research how we can improve the installer. I think there are quite some options for setting e.g. the `PATH` value that the installer can do.

jpcoenen:
This is coming very soon separately from this PR.

jpcoenen:
An MSI installer for `v0.30.0` is now live at the [releases page](https://github.com/secrethub/secrethub-cli/releases/).

Direct download links:
[64 bit / amd64 / x86_x64](https://github.com/secrethub/secrethub-cli/releases/download/v0.30.0/secrethub-v0.30.0-windows-amd64.msi)
[32-bit / i386 / x86](https://github.com/secrethub/secrethub-cli/releases/download/v0.30.0/secrethub-v0.30.0-windows-386.msi)

PR #64, opened 2019-04-12
Add prompt on org init whether free user wants to start trial

florisvdg:


internals/secrethub/org_init.go
```diff
@@ -39,14 +39,30 @@ func (cmd *OrgInitCommand) Register(r Registerer) {
 func (cmd *OrgInitCommand) Run() error {
 	var err error
 
+	client, err := cmd.newClient()
+	if err != nil {
+		return errio.Error(err)
+	}
+
+	user, err := client.Me().GetUser()
+	if user.Username == "Replace this predicate with a check whether user is on free plan" {
```
florisvdg: What do we put here?

internals/secrethub/org_init.go
```diff
@@ -39,14 +39,30 @@ func (cmd *OrgInitCommand) Register(r Registerer) {
 func (cmd *OrgInitCommand) Run() error {
 	var err error
 
+	client, err := cmd.newClient()
+	if err != nil {
+		return errio.Error(err)
+	}
+
+	user, err := client.Me().GetUser()
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

internals/secrethub/org_init.go
SimonBarendse: The plan of the user is not relevant. For users, we offer only one plan: the free developer account.

Each organization has its own plan. So when the same user would create a second org, this would start again on a trial.

Therefore, no predicate is needed here, when creating a new organization, it always starts on a trial.

internals/secrethub/org_init.go
```diff
@@ -39,14 +39,30 @@ func (cmd *OrgInitCommand) Register(r Registerer) {
...
+
 	incompleteInput := cmd.name == "" || cmd.description == ""
 	if cmd.force && incompleteInput {
 		return ErrMissingFlags
 
 	} else if !cmd.force && incompleteInput {
 		fmt.Fprintf(
 			cmd.io.Stdout(),
-			"Before initializing a new organization, I need to know a few things about your organization. "+
+			"Before initializing a new organization, we need to know a few things about your organization. "+
```
SimonBarendse: :+1:

Maybe rephrase it even more. This information is mostly for yourself and for the other users in the organization, not for we/us.

internals/secrethub/org_init.go
```diff
@@ -39,14 +39,27 @@ func (cmd *OrgInitCommand) Register(r Registerer) {
 func (cmd *OrgInitCommand) Run() error {
 	var err error
 
+	if !cmd.force {
+		confirmed, err := ui.AskYesNo(cmd.io, "Creating an organization is not available on the free plan, would you to like to proceed and start a 14-day trial?", ui.DefaultNo)
```
mackenbach: Let's word this positively instead. Also, why would we need to prompt? It's free no strings attached for 14 days, so we can just mention it in the success message right?

Something like:

```
Creating organization...
Organization created. Your 14-day trial starts now. To purchase a subscription for your shared organization workspace, visit https://secrethub.io/pricing
```

SimonBarendse: We could also refer to the getting started on organizations docs instead (these are in the making). These will include more detailed information about working with orgs.

florisvdg: Yes, but you do it only once, thereby making it a 'destructive' action. Maybe the user wants to first discuss it with the rest of the team members on whether the moment is right to take some time and start testing the tool in a team setting. So a tiny little prompt won't hurt here, imo.

florisvdg: And for the wording: 
```
Creating an organization will start a 14-day trial (no strings attached) of the Team plan. Visit https://secrethub.io/pricing for more information. Would you like to proceed?
```

mackenbach: Okay cool, if @SimonBarendse agrees I'm fine with this approach. 

I like how it's relaxed your message is. Well done. A tiny copy nitpick would be to add `free` in there:

`Creating an organization will start a free 14-day trial (no strings attached) of the Team plan. Visit https://secrethub.io/pricing for more information. Would you like to proceed?`

mackenbach: Also, the default should be yes. Definitely yes.

PR #73, opened 2019-05-10
Add a default to secrethub run --template

SimonBarendse:
This makes the default usage simpler.

The simple usecase should not require extra configuration. More complex projects that have multiple tempaltes, can configure them by supplying the flag.

### DIFF
--- a/internals/secrethub/run.go
+++ b/internals/secrethub/run.go
@@ -61,7 +61,7 @@ func (cmd *RunCommand) Register(r Registerer) {
 	clause := r.Command("run", "Runs a program and passes environment variables to it, sourcing values from SecretHub.")
 	clause.Arg("command", "The command to execute").Required().StringsVar(&cmd.command)
 	clause.Flag("envar", "Source an environment variable from a secret at a given path with `NAME=<path>`").Short('e').StringMapVar(&cmd.envar)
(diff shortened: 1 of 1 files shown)


SimonBarendse:
@jpcoenen Good observation! :100: 

I've changed it to only use the template when `os.Stat` does not return an error. This also includes e.g. a permission denied error.

Alternatively, we could check `!os.IsNotExists`. In that case, the command fails when there is no permission to read the file (even if it does not exists).

internals/secrethub/run.go
```diff
@@ -80,6 +80,13 @@ func (cmd *RunCommand) Run() error {
 	}
 	envSources = append(envSources, flagSource)
 
+	if cmd.template == "" {
+		const defaultTemplate = "secrethub-env.yml"
```
mackenbach: Consider that the current practices we recommend in the docs is always suffixing the template with `.tpl`. This breaks from that convention. Now it isn't a change to the status quo, but we should be concious of this.

SimonBarendse: @mackenbach We discussed the name last Friday with the team and preferred to use `.yml` to make it clear and intuitive what the format of the file should be.

> And how do you propose we make it consistent with the current practices?

I'd prefer to use the `.yml` extension in the docs as well, as that makes the format self-explanatory in the docs as well.

PR #74, opened 2019-05-11
Improve the help command

SimonBarendse:


jpcoenen:
Indeed waaay better. 

Do we want the help commands to be printed by default if no sub-command is given? I expected some help when I ran `secrethub org`.

mackenbach:
Oh one tiny thing, if we could change the description of the `service` command to: `Manage non-human service accounts` that would be extra great. Shouldn't block the PR, but as we're touching it already, might as well make it a bit more self explanatory.

SimonBarendse:
> Do we want the help commands to be printed by default if no sub-command is given? I expected some help when I ran secrethub org.

@jpcoenen Good suggestion. This would require a change in kingpin (see https://github.com/alecthomas/kingpin/blob/v2.2.6/app.go#L509) . I'll open an issue upstream.

SimonBarendse:
@mackenbach Putting `generate rand` in the commands section would require a pretty dirty fix. (as `generate` is currently a management command). I'd prefer to drop the `rand` sub-command instead (which we want to do anyways).

SimonBarendse:
> Oh one tiny thing, if we could change the description of the service command to: Manage non-human service accounts that would be extra great.

Aren't service accounts for non-humans by definition? I prefer to keep the help-text as short as possible, to keep it simple.

mackenbach:
> Aren't service accounts for non-humans by definition? I prefer to keep the help-text as short as possible, to keep it simple.

I agree on the goal. It needs to be simple and short, but still understandable. However, `service account` is a bit SecretHub lingo. The concept isn't but the way we call it is (to some extend). 

Therefore, I feel adding the `non-human` part clarifies that it's meant for automation, e.g. in CI or for an app. 

> I'd prefer to drop the rand sub-command instead (which we want to do anyways).

Yes let's do that then. Make sure to put it at least in the release notes and possibly in a release email as well.

SimonBarendse:
> However, service account is a bit SecretHub lingo.

I don't think it is. Found it at least for the following services:
- [AWS](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_users.html#id_users_service_accounts)
- [Google Cloud](https://cloud.google.com/compute/docs/access/service-accounts)
- [Windows](https://docs.microsoft.com/en-us/windows/security/identity-protection/access-control/service-accounts)
- [Kubernetes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/)

I'd say our users are familiar with at least one of those.

PR #78, opened 2019-05-15
Mask secrets on stdout and stderr with secrethub run

jpcoenen:
Add secret masking functionality to `secrethub run`. By default all secrets will be masked in `stdout` and `stderr`. If these values are detected, they are replaced with `<redacted by SecretHub>`.

This can be disabled with the `--no-masking`-flag.

A separate goroutine with `MaskedWriter.Run()` is used to write masked content to the underlying `io.Writer`. This is done to support staggered writes: if a secret is long and is split over multiple calls to `Write()`, we still want to be able to mask these values. Therefore the goroutine with `MaskedWriter.Run()` waits for `Timeout` for any subsequent writes.

## Example
```
$ secrethub write path/to/secret
Please type in the value of the secret, followed by an [ENTER]:

$ secrethub run --envar TEST=path/to/secret -- printenv | grep TEST
TEST=<redacted by SecretHub>
```
## Known "issues"

1. If a very low-entropy secret is masked, it might leak through the context of the output. For example, if a secret has as value `user` and the following text is outputted on stdout:
```
The username is: user
```
The following would be outputted after masking:
```
The <redacted by SecretHub>name is: <redacted by SecretHub>
```
This effectively leaks the value of the secret. However, in practice this should not be a real problem because any real secret should have high enough entropy to not show up in normal text.

2. If someone has control over the process ran with `secrethub run` _and_ the resulting masked logs, an oracle could be setup that leaks any secret by looking at the response time of the log output. However, if someone has access to the ran process, they would also have access to the environment containing the secrets.

### DIFF
--- a/internals/cli/masker/writer.go
+++ b/internals/cli/masker/writer.go
@@ -1,3 +1,4 @@
+// Package masker provides a wrapper around an io.Writer that replaces sensitive values in its output.
 package masker
 
 import (
@@ -6,6 +7,7 @@ import (
 	"time"
 )
 
+// Matcher is an interface used by MaskedWriter to find matches of sequences to mask.
 type Matcher interface {
 	Read(byte) int
 	InProgress() bool
@@ -111,7 +113,7 @@ func NewMaskedWriter(w io.Writer, masks [][]byte, maskString string, timeout tim
 	}
 }
 
-// Write implements implements to Write from io.Writer
+// Write implements to Write from io.Writer
 // It is responsible for finding any matches to mask and mark the appropriate bytes as masked.
 // This function never returns an error. These can instead be caught with Flush().
 func (mw *MaskedWriter) Write(p []byte) (n int, err error) {
@@ -150,7 +152,7 @@ func (mw *MaskedWriter) flushBuffer() {
 	mw.buf = mw.buf[:0]
 }
 
-// Run writes any data processed data from the output channel to the underlying io.Writer.
+// Run writes any processed data from the output channel to the underlying io.Writer.
 // If no new data is received on the output channel for Timeout, the output buffer is forced flushed
 // and all ongoing matches are reset.
 //
--- a/internals/cli/masker/writer.go
+++ b/internals/cli/masker/writer.go
@@ -32,7 +32,7 @@ func (m *sequenceMatcher) Read(in byte) int {
 		return 0
 	}
 
-	m.findShift()
+	m.currentIndex -= m.findShift()
 
 	if m.sequence[m.currentIndex] == in {
 		return m.Read(in)
@@ -44,7 +44,7 @@ func (m *sequenceMatcher) Read(in byte) int {
 // findShift checks whether we can also make a partial match by decreasing the currentIndex .
 // For example, if the sequence is foofoobar, if someone inserts foofoofoobar, we still want to match.
 // So after the third f is inserted, the currentIndex is decreased by 3 with the following code.
-func (m *sequenceMatcher) findShift() {
+func (m *sequenceMatcher) findShift() int {
 	for offset := 1; offset <= m.currentIndex; offset++ {
 		ok := true
 		for i := 0; i < m.currentIndex-offset; i++ {
@@ -54,10 +54,10 @@ func (m *sequenceMatcher) findShift() {
 			}
 		}
 		if ok {
-			m.currentIndex -= offset
-			return
+			return offset
 		}
 	}
+	return m.currentIndex
 }
 
 // InProgress returns whether this sequenceMatcher is currently partially matching.
--- a/internals/cli/masker/writer.go
+++ b/internals/cli/masker/writer.go
@@ -1,0 +1,204 @@
+package masker
+
+import (
+	"io"
+	"sync"
+	"time"
+)
+
+type Matcher interface {
+	Read(byte) int
+	InProgress() bool
+	Reset()
+}
+
+type sequenceMatcher struct {
+	sequence     []byte
+	currentIndex int
+}
+
+// Read takes in a new byte to match against.
+// If the given byte results in a match with sequence, the number of matched bytes is returned.
+func (m *sequenceMatcher) Read(in byte) int {
+	if m.sequence[m.currentIndex] == in {
+		m.currentIndex++
+
+		if m.currentIndex == len(m.sequence) {
+			m.currentIndex = 0
+			return len(m.sequence)
+		}
+		return 0
+	}
+
+	m.findShift()
+
+	if m.sequence[m.currentIndex] == in {
+		return m.Read(in)
+	}
+	m.currentIndex = 0
+	return 0
+}
+
+// findShift checks whether we can also make a partial match by decreasing the currentIndex .
+// For example, if the sequence is foofoobar, if someone inserts foofoofoobar, we still want to match.
+// So after the third f is inserted, the currentIndex is decreased by 3 with the following code.
+func (m *sequenceMatcher) findShift() {
+	for offset := 1; offset <= m.currentIndex; offset++ {
+		ok := true
+		for i := 0; i < m.currentIndex-offset; i++ {
+			if m.sequence[i] != m.sequence[i+offset] {
+				ok = false
+				break
+			}
+		}
+		if ok {
+			m.currentIndex -= offset
+			return
+		}
+	}
+}
+
+// InProgress returns whether this sequenceMatcher is currently partially matching.
+//
+// For example, if the sequence is "foobar" and the registered input is "foob", InProgress() returns true.
+func (m *sequenceMatcher) InProgress() bool {
+	return m.currentIndex > 0
+}
+
+// Reset forgets the current match.
+func (m *sequenceMatcher) Reset() {
+	m.currentIndex = 0
+}
+
+// maskByte represents a byte and whether the byte should be masked or not.
+type maskByte struct {
+	byte
+	masked bool
+}
+
+// MaskedWriter wraps an io.Writer and masks all matches found by Matchers with MaskString.
+// If no write is made for Timeout on the io.Writer, any matches in progress are reset
+// and the buffer is flushed. This is to ensure that the writer does not hang on partial matches.
+type MaskedWriter struct {
+	w          io.Writer
+	MaskString string
+	Matchers   []Matcher
+	Timeout    time.Duration
+
+	buf    []maskByte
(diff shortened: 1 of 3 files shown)


internals/cli/masker/writer.go
```diff
@@ -0,0 +1,204 @@
...
+	copy(tmp, mw.buf)
+	mw.output <- tmp
+	mw.buf = mw.buf[:0]
+}
+
+// Run writes any data processed data from the output channel to the underlying io.Writer.
+// If no new data is received on the output channel for Timeout, the output buffer is forced flushed
+// and all ongoing matches are reset.
+//
+// This should be run in a seperate goroutine.
```
golangcibot: `seperate` is a misspelling of `separate` (from `misspell`)

internals/cli/masker/writer.go
```diff
@@ -0,0 +1,204 @@
...
+			for i := 0; i < maskLen; i++ {
+				mw.buf[len(mw.buf)-1-i].masked = true
+			}
+			matchInProgress = matchInProgress || matcher.InProgress()
+		}
+
+		if !matchInProgress {
+			mw.flushBuffer()
+		} else {
+			mw.output <- []maskByte{}
```
SimonBarendse: What's this for?

internals/cli/masker/writer.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func (mw *MaskedWriter) flushBuffer() {
+	tmp := make([]maskByte, len(mw.buf))
+	copy(tmp, mw.buf)
+	mw.output <- tmp
+	mw.buf = mw.buf[:0]
+}
+
+// Run writes any data processed data from the output channel to the underlying io.Writer.
```
SimonBarendse: `any data processed data`

typo?

internals/cli/masker/writer.go
```diff
@@ -0,0 +1,204 @@
...
+// If no new data is received on the output channel for Timeout, the output buffer is forced flushed
+// and all ongoing matches are reset.
+//
+// This should be run in a separate goroutine.
```
SimonBarendse: If this should always be run in a separate goroutine, does is make sense to start that goroutine in this function?

internals/secrethub/run.go
```diff
@@ -37,15 +39,21 @@ var (
 	ErrEnvFileFormat  = errRun.Code("invalid_env_file_format").ErrorPref("env-file templates must be a valid yaml file with a map of string key and value pairs: %v")
 )
 
+var (
```
SimonBarendse: This can be a `const`

internals/secrethub/run.go
```diff
@@ -198,12 +226,23 @@ func (cmd *RunCommand) Run() error {
...
 
-	if err != nil {
+	if !cmd.noMasking {
+		err = maskedStdout.Flush()
+		if err != nil {
+			fmt.Println(err)
+		}
+		err = maskedStderr.Flush()
+		if err != nil {
+			fmt.Println(err)
```
SimonBarendse: Shouldn't this be written to stderr?

internals/cli/masker/writer.go
```diff
@@ -0,0 +1,204 @@
+package masker
```
SimonBarendse: Could you add a short package description (godoc)?

Optionally add a larger description as well.

internals/cli/masker/writer.go
```diff
@@ -0,0 +1,204 @@
...
+
+import (
+	"io"
+	"sync"
+	"time"
+)
+
+type Matcher interface {
```
SimonBarendse: Please add some docs to this exported interface.

internals/cli/masker/writer.go
```diff
@@ -0,0 +1,204 @@
...
+	for offset := 1; offset <= m.currentIndex; offset++ {
+		ok := true
+		for i := 0; i < m.currentIndex-offset; i++ {
+			if m.sequence[i] != m.sequence[i+offset] {
+				ok = false
+				break
+			}
+		}
+		if ok {
+			m.currentIndex -= offset
```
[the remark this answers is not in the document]
jpcoenen: That makes a lot of sense :ok_hand:

internals/cli/masker/writer.go
```diff
@@ -0,0 +1,204 @@
...
+		w:          w,
+		MaskString: maskString,
+		Matchers:   matchers,
+		Timeout:    timeout,
+		lock:       &lock,
+		output:     make(chan []maskByte, 1),
+	}
+}
+
+// Write implements implements to Write from io.Writer
```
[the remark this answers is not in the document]
jpcoenen: That sure is a typo

internals/cli/masker/writer.go
jpcoenen: To reset the timeout. This way every call to `Write` resets the timeout. But thinking of it, it might make more sense not to do this. Then the timeout actually defines the maximum time between two flushes.

internals/cli/masker/writer.go
jpcoenen: I love them duplicated words duplicated.

internals/cli/masker/writer.go
jpcoenen: As far as I know, idiomatic Go prescribes that you should leave this to the caller. Let me look that up.

jpcoenen: https://talks.golang.org/2013/bestpractices.slide#25
https://stackoverflow.com/questions/28506013/idiomatic-golang-goroutines
https://hackmongo.com/page/golang-antipatterns/#hidden-goroutine

internals/cli/masker/writer.go
jpcoenen: Changed it.

internals/cli/masker/writer.go
SimonBarendse: Let's keep it as is :+1:

internals/cli/masker/writer.go
SimonBarendse: Could you document the functions as well?

internals/cli/masker/writer.go
```diff
@@ -32,7 +32,7 @@ func (m *sequenceMatcher) Read(in byte) int {
 		return 0
 	}
 
-	m.findShift()
+	m.currentIndex -= m.findShift()
```
SimonBarendse: Better! :+1: 

I have one more proposal, how about:

```Go
shift := m.findShift()
if shift > 0 {
    m.currentIndex -= shift
    return m.Read(in)
}
m.currentIndex = 0
return 0
```

I think this makes it clearer when the if statement is true.

internals/secrethub/run.go
```diff
@@ -232,11 +232,11 @@ func (cmd *RunCommand) Run() error {
 	if !cmd.noMasking {
 		err = maskedStdout.Flush()
 		if err != nil {
-			fmt.Println(err)
+			fmt.Fprintln(os.Stderr, err)
```
SimonBarendse: `os.Stdout`?

internals/cli/masker/writer.go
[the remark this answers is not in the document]
SimonBarendse: Should `to` be in there? Still doesn't feel very fluent.

internals/cli/masker/writer.go
SimonBarendse: document-all-things

jpcoenen: Or shall I just privatize the whole interface? There is no reason to export it anyway.

internals/cli/masker/writer.go
jpcoenen: Shift is always `>0`.

internals/cli/masker/writer.go
SimonBarendse: Currently, you can create your own implementation of `Matcher` and construct an instance of `MaskedWriter` which uses this own implementation.

We could make it private. If we do, `MaskedWriter.Matchers` should be made private as well.
It is then no longer possible to create a valid `MaskedWriter` using the struct outside the package. Do we want to make the other fields private as well then?

I'd prefer to keep it as is and document the functions.

internals/cli/masker/writer.go
jpcoenen: Made it a bit simpler though

internals/secrethub/run.go
jpcoenen: Why? It's an error.

internals/cli/masker/writer.go
SimonBarendse: Ahh, that was related to https://github.com/secrethub/secrethub-cli/commit/d56093a0680553b739de6d1da502219bde8fd8e2#r33578665

SimonBarendse: Then we can drop the `m.currentIndex = 0` after the if statement, right? Because when the if statement is not true, shift was `m.currentIndex`, so the new index became 0 after substracting the shift.

internals/secrethub/run.go
SimonBarendse: Ahh nvm

internals/cli/masker/writer.go
jpcoenen: You'll still be able to do that if it's private. As long as your implementation satisfies the interface, you can make it work. If it's export it means you can use this interface outside of the package. Meaning that it would be useful for someone to use this interface. I don't think it is, therefore it would be better to make it private.

internals/cli/masker/writer.go
jpcoenen: Yeah. I came to the same conclusion.

internals/cli/masker/writer.go
```diff
@@ -0,0 +1,209 @@
...
+	buf    []maskByte
+	lock   *sync.Mutex
+	output chan []maskByte
+	err    error
+	nIn    int64
+	nOut   int64
+}
+
+// NewMaskedWriter returns a MaskedWriter which masks all occurrences of sequences in masks from the underlyting io.Writer.
+func NewMaskedWriter(w io.Writer, masks [][]byte, maskString string, timeout time.Duration) *maskedWriter {
```
golangcibot: exported func NewMaskedWriter returns unexported type *masker.maskedWriter, which can be annoying to use (from `golint`)

internals/cli/masker/writer.go
SimonBarendse: godoc is really clear now :+1:

PR #82, opened 2019-05-27
Add org purchase command

SimonBarendse:


internals/secrethub/org.go
```diff
@@ -22,6 +22,7 @@ func (cmd *OrgCommand) Register(r Registerer) {
 	NewOrgInitCommand(cmd.io, cmd.newClient).Register(clause)
 	NewOrgInspectCommand(cmd.io, cmd.newClient).Register(clause)
 	NewOrgInviteCommand(cmd.io, cmd.newClient).Register(clause)
+	NewOrgPurchaseCommand(cmd.io).Register(clause)
```
florisvdg: Should it be on the `org` level? Could also just be `secrethub purchase`, just like `secrethub signup`.

SimonBarendse: A subscription is purchased by and for an organization. Users that do not use organizations don't have to purchase, as personal accounts are free. Therefore, I think the purchase command should be a subcommand of `org`. This also keeps the `--help` output relevant. `purchase` is not relevant for users that do not use organizations.

florisvdg: Okay, though we could it the future make more non-org features premium as well. But since you won't script this command anyway, it's fine to break it if we ever decide to do that.

SimonBarendse: I don't think that's the intention.

Even if it were, we can add a `purchase` command at root level later and add a (hidden) alias from this command. No need to break :smile:

PR #91, opened 2019-06-07
 Add support for template vars to the run command 

SimonBarendse:
This work continued on the work of #90 .

Please address that PR before reviewing this one.

### DIFF
--- a/internals/secrethub/inject.go
+++ b/internals/secrethub/inject.go
@@ -11,7 +11,7 @@ import (
 	"github.com/secrethub/secrethub-cli/internals/cli/filemode"
 	"github.com/secrethub/secrethub-cli/internals/cli/posix"
 	"github.com/secrethub/secrethub-cli/internals/cli/ui"
-	"github.com/secrethub/secrethub-cli/internals/tpl"
+	"github.com/secrethub/secrethub-cli/internals/secrethub/tpl"
 
 	"github.com/secrethub/secrethub-go/internals/errio"
 	"github.com/secrethub/secrethub-go/pkg/secrethub"
@@ -29,6 +29,7 @@ type InjectCommand struct {
 	clearClipboardAfter time.Duration
 	clipper             clip.Clipper
 	newClient           newClientFunc
+	templateVars        map[string]string
 }
 
 // NewInjectCommand creates a new InjectCommand.
@@ -38,6 +39,7 @@ func NewInjectCommand(io ui.IO, newClient newClientFunc) *InjectCommand {
 		clearClipboardAfter: defaultClearClipboardAfter,
 		io:                  io,
 		newClient:           newClient,
+		templateVars:        make(map[string]string),
 	}
 }
 
@@ -76,15 +78,20 @@ func (cmd *InjectCommand) Run() error {
 		return errio.Error(err)
 	}
 
-	tpl, err := tpl.NewParser("${", "}").Parse(string(raw))
+	varTemplate, err := tpl.NewV1Parser().Parse(string(raw))
 	if err != nil {
 		return errio.Error(err)
+	}
+
+	secretTemplate, err := varTemplate.InjectVars(cmd.templateVars)
+	if err != nil {
+		return err
 	}
 
 	secrets := make(map[string]string)
 
 	var client secrethub.Client
-	secretPaths := tpl.Keys()
+	secretPaths := secretTemplate.Secrets()
 	if len(secretPaths) > 0 {
 		client, err = cmd.newClient()
 		if err != nil {
@@ -100,7 +107,7 @@ func (cmd *InjectCommand) Run() error {
 		secrets[path] = string(secret.Data)
 	}
 
-	injected, err := tpl.Inject(secrets)
+	injected, err := secretTemplate.InjectSecrets(secrets)
 	if err != nil {
 		return errio.Error(err)
 	}
--- a/internals/secrethub/run.go
+++ b/internals/secrethub/run.go
@@ -70,7 +70,7 @@ func (cmd *RunCommand) Register(r Registerer) {
 	clause.Arg("command", "The command to execute").Required().StringsVar(&cmd.command)
 	clause.Flag("envar", "Source an environment variable from a secret at a given path with `NAME=<path>`").Short('e').StringMapVar(&cmd.envar)
 	clause.Flag("template", "The path to a .yml template file with environment variable mappings of the form `NAME: value`. Templates are automatically injected with secrets when referenced.").StringVar(&cmd.template)
-	clause.Flag("var", "Set variables to be used in templates.").StringMapVar(&cmd.templateVars)
+	clause.Flag("var", "Set variables to be used in templates.").Short('v').StringMapVar(&cmd.templateVars)
 	clause.Flag("env", "The name of the environment prepared by the set command (default is `default`)").Default("default").Hidden().StringVar(&cmd.env)
 	clause.Flag("no-masking", "Disable masking of secrets on stdout and stderr").BoolVar(&cmd.noMasking)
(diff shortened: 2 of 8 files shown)


internals/secrethub/tpl/template_test.go
```diff
@@ -0,0 +1,74 @@
+package tpl_test
+
+import (
+	"testing"
+
+	"github.com/secrethub/secrethub-cli/internals/secrethub/tpl"
+	"github.com/secrethub/secrethub-go/internals/assert"
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

internals/secrethub/run.go
```diff
@@ -320,13 +324,47 @@ func (t envTemplate) Secrets() []string {
 	return result
 }
 
+type envTemplate struct {
+	envVars map[string]secrethubtpl.SecretTemplate
+}
+
+func (t envTemplate) Env(secrets map[string]string) (map[string]string, error) {
```
jpcoenen: Missing a comment here

internals/secrethub/tpl/template.go
```diff
@@ -0,0 +1,106 @@
...
+
+// Template is a template containing secret paths.
+//
+// Secret paths can be supplied between brackets:
+// {{ path/to/secret }}.
+// Secret paths can contain variables. These can be
+// supplied using: ${ var }.
+// Combined this can look like:
+// {{ ${app}/db/secret }}
+type Template struct {
```
jpcoenen: I found myself being pretty confused about the name of this. Especially because this `Template` contains a map of other `Template` structs. Is there a way in which we can distinguish between the two? Or is there a specific reason for this naming?

SimonBarendse: `internals/tpl` is very generic; It allows parsing and substitution of any value between any delimiter. 
`internals/secrethub/tpl` is more specific; It allows parsing of secret paths given by `{{` `}}` delimiters, which can contain variables given by `${` `}` delimiters.

`internals/secrethub/tpl` is not specifically for run templates, it can be used to inject files as well.

So you could say the `internals/secrethub/tpl` is specifically a SecretHub template, while `internals/tpl` can be any template.

---

Having this extra information, what would be a name that explains this best? Should we put more emphasis on the `secrethub` import path of `internals/secrethub/tpl`?

internals/secrethub/tpl/v2.go
```diff
@@ -0,0 +1,113 @@
+package tpl
+
+import (
+	"fmt"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

internals/secrethub/run_test.go
```diff
@@ -4,52 +4,52 @@ import (
...
+					value:      "{{path/to/secret}}",
+					lineNumber: 2,
+				},
 			},
 		},
 		"= sign in value": {
 			raw: "foo=foo=bar",
-			expected: map[string]string{
-				"foo": "foo=bar",
+			expected:[]envvar{
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

internals/secrethub/run.go
```diff
@@ -286,14 +289,16 @@ type EnvSource interface {
 	Secrets() []string
 }
 
-type envTemplate struct {
-	vars map[string]tpl.Template
+type ymlTemplate struct {
```
golangcibot: U1000: type `ymlTemplate` is unused (from `unused`)

internals/secrethub/tpl/v2.go
```diff
@@ -0,0 +1,113 @@
...
+// NewV2Parser returns a parser for the v2 template syntax.
+//
+// V2 templates can contain secret paths between brackets:
+// {{ path/to/secret }}
+//
+// Secret paths can contain variables between ${}:
+// ${ var }
+//
+// Combined this can look like:
+// {{ ${app}/db/secret }}
```
mackenbach: Does the `v2` syntax support variables outside of brackets? E.g. something like this:

```
DB_USER = ${env}-user
DB_PASSWORD = {{ company/repo/${env}/app/db_password }}
```

When we talk about writing a spec for the syntax, this is one of the things we mean. Consider expanding this GoDoc comment to a bit more complete description of the syntax and what behavior people can expect (e.g. are spaces ignored in vars `${ foo }` and in path substitutions `{{ foo }}`. 

Then we can later create more complete user documentation from the code documentation.

internals/secrethub/inject.go
```diff
@@ -76,15 +78,20 @@ func (cmd *InjectCommand) Run() error {
 		return errio.Error(err)
 	}
 
-	tpl, err := tpl.NewParser("${", "}").Parse(string(raw))
+	varTemplate, err := tpl.NewV1Parser().Parse(string(raw))
```
mackenbach: I'm missing something. Didn't we discuss that `v2` would be the default syntax except for `.yml` formatted env files? It seems to me now like `v1` is not only the default but also the only supported one.

internals/secrethub/run.go
```diff
@@ -400,31 +400,68 @@ func (e EnvFile) Secrets() []string {
...
-		}
-		return template, nil
+		return nil, err
 	}
 
-	secretTemplates := make(map[string]tpl.SecretTemplate, len(templates))
-	for k, template := range templates {
+	secretTemplates := make(map[string]tpl.SecretTemplate, len(env))
+	for _, envvar := range env {
+		err = validation.ValidateEnvarName(envvar.key)
```
mackenbach: This has to be `Posix` right?

internals/secrethub/run.go
```diff
@@ -438,55 +475,41 @@ func parseEnv(raw string) (map[string]tpl.VarTemplate, error) {
...
-		}
-
-		err = validation.ValidateEnvarName(key)
-		if err != nil {
-			return nil, ErrTemplate(i, err)
-		}
+		vars[key] = envvar{
+			key:        key,
+			value:      value,
+			lineNumber: i}
```
mackenbach: Tiny, tiny thing, but if you're going to do multiline struct declarations, which I actually like, then please put the closing bracket on a new line as well for readability:

```
   lineNumber: i,
}
```

internals/secrethub/inject.go
SimonBarendse: This PR does not change the inject command. The v2 syntax and variable support still has to be added to the inject command.
As this PR is already pretty big, I propose to add this in a separate PR.

SimonBarendse: This PR also does not add the discussed `--template-version` flag to the run and inject commands.

internals/secrethub/run.go
SimonBarendse: This hasn't been changed (https://github.com/secrethub/secrethub-cli/blob/v0.23.0/internals/secrethub/run.go#L346). Do we want it to?

internals/secrethub/tpl/v2.go
SimonBarendse: > Does the v2 syntax support variables outside of brackets?

It does not. I've updated the godoc to explicitly say this.

Spaces directly after opening delimiters and directly before closing delimiters are ignored. I've added a comment about this to the godoc.

jpcoenen: > > Does the v2 syntax support variables outside of brackets?  

> It does not. I've updated the godoc to explicitly say this.

Isn't this something we want? Can't we easily solve this by swapping the two template parsers around? So first variable substitution and _then_ secrets?

mackenbach: Yes that makes absolute sense to me. First do variable substition and then do secret injection.

internals/secrethub/run.go
mackenbach: Hmm ok, I don't think we need to. Though not entirely sure what the impact is for the `KEY=VALUE` pattern.

PR #94, opened 2019-06-07
Use secrethub.env run template when it exists and no other template is supplied

SimonBarendse:


internals/secrethub/run.go
```diff
@@ -87,6 +87,13 @@ func (cmd *RunCommand) Run() error {
 	}
 	envSources = append(envSources, flagSource)
 
+	if cmd.template == "" {
+		const defaultTemplate = "secrethub.env"
+		if _, err := os.Stat(defaultTemplate); err == nil {
```
florisvdg: Probably safer like this? Because if it does exist and there is an error, you do want to know about it.
```go
if _, err := os.Stat(defaultTemplate); os.IsNotExist(err) {
   cmd.template = defaultTemplate
}
```

internals/secrethub/run.go
```diff
@@ -89,7 +89,12 @@ func (cmd *RunCommand) Run() error {
 
 	if cmd.template == "" {
 		const defaultTemplate = "secrethub.env"
-		if _, err := os.Stat(defaultTemplate); err == nil {
+		_, err := os.Stat(defaultTemplate)
+		if err != nil {
+			if !os.IsNotExist(err) {
```
mackenbach: This is what the `&&` operator was invented for. Let's try to avoid these nested if statements if we can keep it simple.

SimonBarendse: Combining with `&&` breaks the `else` statement below, which only belongs to the `err != nil` check.

SimonBarendse: These are the three cases:
err != nil && !os.IsnotExists(err) => log
err != nil && os.IsnotExists(err) => ignore
err == nil => use default template

mackenbach: Yes that's what I meant, making it explicit like this improves the readability of the code. It's not blocking, but something that caught my eye.

SimonBarendse: Oke interesting

You'd duplicate the `err != nil` check (both in code and execution) so that the condition's are together on one line, which improves readability.

Might be interesting to dig in some more (and write a blogpost about, don't see a lot of this out there yet).

For example:
```go
if conditionA && conditionB {
  // Handle
}
if conditionA && !conditionB {
  // Handle
}
if !conditionA {
  // Handle
}
```
vs
```go
if conditionA {
  if conditionB {
    // Handle
  } else {
    // Handle
  }
} else {
  // Handle
}
```

The former has the checks together on one line which improves readability, at the cost of three extra condition checks.
The latter guarantees that the checks are exhaustive, i.e. either of the codeblocks is executed. In the first example either of the codeblocks is executed as well, but this is not guaranteed by an `else`. Where in the first example a bug in a condition could mean none or two codeblocks are executed, the second example ensures that always one and only one of the codeblocks is executed. That might be something to consider in some usecases. And maybe more importantly: it's easier to see at one glance that the checks are exhaustive.

SimonBarendse: Oke interesting

You'd duplicate the `err != nil` check (both in code and execution) so that the condition's are together on one line, which improves readability.

Might be interesting to dig in some more (and write a blogpost about, don't see a lot of this out there yet).

For example:
```go
if conditionA && conditionB {
  // Handle
}
if conditionA && !conditionB {
  // Handle
}
if !conditionA {
  // Handle
}
```
vs
```go
if conditionA {
  if conditionB {
    // Handle
  } else {
    // Handle
  }
} else {
  // Handle
}
```

The former has the checks together on one line which improves readability, at the cost of three extra condition checks.
In the latter it's easier to see at a glance that checks are exhaustive, i.e. either of the codeblocks is executed. Might be important in some usecases.

PR #95, opened 2019-06-07
Allow extra spaces around = sign in .env run templates

SimonBarendse:


florisvdg:
Would be nice to have a unit test that proves this.

PR #98, opened 2019-06-13
Use v2 template syntax in the inject command

SimonBarendse:


internals/secrethub/inject.go
```diff
@@ -19,6 +21,11 @@ import (
 	"github.com/docker/go-units"
 )
 
+// Errors
+var (
+	ErrUnknownTemplateVersion = errMain.Code("unknown_template_version").ErrorPref("unkown template version: '%s' supported versions are 1, 2 and latest")
```
golangcibot: `unkown` is a misspelling of `unknown` (from `misspell`)

internals/secrethub/inject.go
```diff
@@ -78,7 +85,19 @@ func (cmd *InjectCommand) Run() error {
...
 
-	varTemplate, err := tpl.NewV1Parser().Parse(string(raw))
+	var parser tpl.Parser
+	switch cmd.templateVersion {
+	case "1":
+		parser = tpl.NewV1Parser()
+	case "2":
+		parser = tpl.NewV2Parser()
+	case "latest":
+		parser = tpl.NewParser()
```
mackenbach: 😍

internals/secrethub/inject.go
```diff
@@ -62,6 +62,7 @@ func (cmd *InjectCommand) Register(r Registerer) {
 	).Short('c').BoolVar(&cmd.useClipboard)
 	clause.Flag("file", "Write the injected template to a file instead of stdout.").StringVar(&cmd.file)
 	clause.Flag("file-mode", "Set filemode for the file if it does not yet exist. Defaults to 0600 (read and write for current user) and is ignored without the --file flag.").Default("0600").SetValue(&cmd.fileMode)
+	clause.Flag("var", "Set variables to be used in templates.").Short('v').StringMapVar(&cmd.templateVars)
```
mackenbach: Small thing: I don't really like the word `set`. It is rarely clear what `set` does. Something like this is maybe more clear:

`Define the value for a template variable with VAR=VALUE, e.g. --var env=prod`

internals/secrethub/inject.go
```diff
@@ -78,12 +88,48 @@ func (cmd *InjectCommand) Run() error {
 		return errio.Error(err)
 	}
 
-	varTemplate, err := tpl.NewV1Parser().Parse(string(raw))
+	templateVars := make(map[string]string)
```
mackenbach: I feel this piece of code up until the posix validation may be duplicated from the run command. Consider making a small helper function for it. 

Also, can we skip variable reading when we're using `v1`? Not super important as performance isn't a big factor here so if doing it anyway keeps it simpler stick to that. Only refactor if it makes more sense that way.

internals/secrethub/inject.go
mackenbach: Also, are variables case sensitive? If so, state it in the docs.

IMO, they should be case *insensitive*.

jpcoenen:
Found a buggy (I think):
```
$ echo "{{joriscoenen/filetest/\${hoi}}}" | ./secrethub inject --var hoi=aap 
Encountered an error: missing closing delimiter '}' (template.tag_not_closed)
```

internals/secrethub/inject.go
SimonBarendse: > Also, can we skip variable reading when we're using v1?

Currently, we're returning an error when template vars are supplied. This also happens when template vars are supplied via envvars. So, reading the vars should also be done in `v1` to  check whether no var is set.

mackenbach:
@SimonBarendse are you planning to fix the bug Joris found above? https://github.com/secrethub/secrethub-cli/pull/98#issuecomment-501726030

SimonBarendse:
> @SimonBarendse are you planning to fix the bug Joris found above? #98 (comment)

Yes. That should be fixed before releasing imo.

This is a bug in the v2 syntax btw, so it also affects the run command.

PR #99, opened 2019-06-13
Add comment support to run templates

SimonBarendse:


internals/secrethub/run.go
```diff
@@ -460,9 +460,15 @@ func parseEnv(raw string) ([]envvar, error) {
 	vars := map[string]envvar{}
 	scanner := bufio.NewScanner(strings.NewReader(raw))
 
-	i := 1
+	i := 0
 	for scanner.Scan() {
+		i++
 		line := scanner.Text()
+
+		if line == "" || strings.HasPrefix(line, "#") {
```
mackenbach: What happens when I have a line with only whitespace?

This could be a bit more foolproof solution:
```
if len(strings.TrimSpace(line)) == 0 || strings.HasPrefix(line, "#") {
   ...
}

Comment (SimonBarendse) [reply internals/secrethub/run.go]: > What happens when I have a line with only whitespace?

Why would we want to allow this?

Comment (jpcoenen) [reply internals/secrethub/run.go]: Because in the user's experience it is the same as an empty line. I'd even go as far as saying that `     # comment` is still a comment. So then it'd become:
```
trimmed := string.TrimSpace(line)
if len(trimmed) == 0 || strings.HasPrefix(trimmed, "#") {
   ...
}
```

SimonBarendse:
> pick up the other todos in separate PRs

Hi @mackenbach , what todos from this PR are still open? I think everything mentioned has been handled.

mackenbach:
@SimonBarendse we discussed yesterday that the .ENV file needed some work, e.g. adding the ability to wrap values in quotes, so that's what I meant with other TODOs. This PR was about adding comment support, which is finished.

PR #100, opened 2019-06-19
Add quote support to .ENV files

mackenbach:
This makes the support similar to the [Node `dotenv` project](https://www.npmjs.com/package/dotenv#rules), which a lot of people are used to. The only feature we don't yet have is multiline support in quoted values, but we've decided to leave that for later. 

Also something to consider is how to handle quoted keys in the KEY=VALUE syntax. We can either pass them as is to the environment or ignore them and filter them out. IMO, passing them as is would be the simplest solution.

### DIFF
--- a/internals/secrethub/run.go
+++ b/internals/secrethub/run.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"syscall"
 	"time"
+	"unicode"
 
 	"github.com/secrethub/secrethub-cli/internals/cli/masker"
 	"github.com/secrethub/secrethub-cli/internals/cli/validation"
@@ -469,14 +470,25 @@ func parseDotEnv(r io.Reader) ([]envvar, error) {
 			return nil, ErrTemplate(i, errors.New("template is not formatted as key=value pairs"))
 		}
 
+		valColNo := len(parts[0]) + 2 // the length of the key (including spaces and quotes) + one for the = sign and one for the current column.
+		for _, r := range parts[1] {
+			if !unicode.IsSpace(r) {
+				break
+			}
+			valColNo++
+		}
+
 		key := strings.TrimSpace(parts[0])
-		value := trimQuotes(strings.TrimSpace(parts[1]))
+		value, isTrimmed := trimQuotes(strings.TrimSpace(parts[1]))
+		if isTrimmed {
+			valColNo++
+		}
 
 		vars[key] = envvar{
 			key:          key,
 			value:        value,
 			lineNumber:   i,
-			columnNumber: len(parts[0]) + 1 + len(parts[1]) - len(value) + 1, // the length of the key (including extra spaces) + the length of the = char + the length of the spaces before the value + 1 for the current char
+			columnNumber: valColNo,
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


internals/secrethub/run_test.go
```diff
@@ -72,6 +72,16 @@ func TestParseEnv(t *testing.T) {
...
+		"success with tabs": {
+			raw: "key\t=\tvalue",
+			expected: []envvar{
+				{
+					key:        "key",
+					value:      "value",
+					lineNumber: 1,
+				},
+			},
+		},
```
SimonBarendse: Can you add a testcase for white space before the key and one for white space after the value?

internals/secrethub/run.go
```diff
@@ -498,6 +498,32 @@ func parseEnv(raw string) ([]envvar, error) {
 	return res, nil
 }
 
+const (
+	doubleQuoteChar = '\u0022'
```
SimonBarendse: why not just `"` and `\'`? Those would be easier to read.

internals/secrethub/run_test.go
```diff
@@ -82,6 +82,36 @@ func TestParseEnv(t *testing.T) {
...
+		"success with quotes and whitespace": {
+			raw: "key = 'value'",
+			expected: []envvar{
+				{
+					key:        "key",
+					value:      "value",
+					lineNumber: 1,
+				},
+			},
+		},
```
SimonBarendse: Could you add a testcase for maintaining white space within quotes?

SimonBarendse: Could you add a testcase for maintaining quotes within the value (e.g. `key={"foo":"bar"}`)?

SimonBarendse:
> Also something to consider is how to handle quoted keys in the KEY=VALUE syntax. We can either pass them as is to the environment or ignore them and filter them out. IMO, passing them as is would be the simplest solution.

Passing them as is means that an error will be thrown, as `"KEY"` does not validate as an environment variable.

I agree that this is the desired behavior. In the dotenv project, quoted keys are skipped. As we have chosen to instead error on invalid input, it makes sense to do so in this case as well.

internals/secrethub/run.go
mackenbach: Chose to use the unicode character to be explicit about which kind of quote we mean. There are quite a lot of different quotation marks in Unicode and we just mean the basic ones, so that's why I wanted to be explicit about it. The name also describes it already so I feel it's pretty clear as is. 

See: https://www.cl.cam.ac.uk/~mgk25/ucs/quotes.html

internals/secrethub/run_test.go
mackenbach: That's already unit tested in `TestTrimQuotes`

mackenbach:
> Passing them as is means that an error will be thrown, as "KEY" does not validate as an environment variable.

I don't think it does error. Tried it out and `"key"=value` evaluates to a quoted `"key"` and `value` and this seems to return `nil`: 

```golang
err := validation.ValidateEnvarName(`"key"`)
fmt.Println(err)
// Prints: <nil>
```

So in that case, we're not modifying it, which I think is the simplest way to go. Sometimes, people need to set weird environment variables, I know Java sometimes wants envars with a dot in them, so we should let them. This also the opinion of the IEEE specification by the way.

internals/secrethub/run.go
SimonBarendse: How is this more explicit than setting a const to the rune itself?

```Go
const doubleQuoteChar = '"'
```
---
`'"' == '\u0022'`, see [this playground](https://goplay.space/#f8-DolKY9MD)

SimonBarendse:
Ahh nvm, `"` are not allowed in Posix envvars, but we validate the envvars on the IEEE spec. :+1:

PR #101, opened 2019-06-19
Rework v2 secret templates

SimonBarendse:

look for variables.

number.

brackets later ($var).

### DIFF
--- a/internals/secrethub/inject.go
+++ b/internals/secrethub/inject.go
@@ -16,7 +16,6 @@ import (
 	"github.com/secrethub/secrethub-cli/internals/secrethub/tpl"
 
 	"github.com/secrethub/secrethub-go/internals/errio"
-	"github.com/secrethub/secrethub-go/pkg/secrethub"
 
 	"github.com/docker/go-units"
 )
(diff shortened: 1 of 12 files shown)


internals/secrethub/tpl/v2_test.go
```diff
@@ -1,23 +1,469 @@
...
+			input: "{{ path/with/${var@b} }}",
+			err:   ErrIllegalVariableCharacter('@', 1, 19),
+		},
+		"error on new line": {
+			input: "{{ path/to/secret }}\n{{ a%b }}",
+			err:   ErrIllegalSecretCharacter('%', 2, 5),
+		},
+		"secret tag not closed": {
+			input: "{{ path",
+			err: ErrSecretTagNotClosed,
```
golangcibot: cannot use ErrSecretTagNotClosed (variable of type func(args ...interface{}) errio.PublicError) as error value in struct literal: missing method Error (from `typecheck`)

internals/secrethub/tpl/v2_test.go
```diff
@@ -1,23 +1,469 @@
...
+			input: "{{ foo/${var}",
+			err: ErrSecretTagNotClosed,
+		},
+		"secret tag not closed after space after var": {
+			input: "{{ foo/${var} ",
+			err: ErrSecretTagNotClosed,
+		},
+		"variable tag not closed": {
+			input: "${ var",
+			err: ErrVariableTagNotClosed,
```
golangcibot: cannot use ErrVariableTagNotClosed (variable of type func(args ...interface{}) errio.PublicError) as error value in struct literal: missing method Error (from `typecheck`)

internals/secrethub/inject.go
```diff
@@ -124,36 +123,17 @@ func (cmd *InjectCommand) Run() error {
 		return ErrUnknownTemplateVersion(cmd.templateVersion)
 	}
 
-	varTemplate, err := parser.Parse(string(raw))
+	template, err := parser.Parse(string(raw), 1,1)
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

SimonBarendse:
Opened #103 to fix this.

PR #102, opened 2019-06-19
Rework v2 secret templates

SimonBarendse:

look for variables.

number.

brackets later ($var).


SimonBarendse:
Opened #103 to fix this

PR #103, opened 2019-06-19
Rework v2 secret templates

SimonBarendse:

look for variables.

number.

brackets later ($var).

### DIFF
--- a/internals/secrethub/run.go
+++ b/internals/secrethub/run.go
@@ -2,8 +2,10 @@ package secrethub
 
 import (
 	"bufio"
+	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -352,11 +354,11 @@ func (t envTemplate) Secrets() []string {
 
 // ReadEnvFile reads and parses a .env file.
 func ReadEnvFile(filepath string, vars map[string]string) (EnvFile, error) {
-	content, err := ioutil.ReadFile(filepath)
+	r, err := os.Open(filepath)
 	if err != nil {
 		return EnvFile{}, ErrCannotReadFile(filepath, err)
 	}
-	env, err := NewEnv(string(content), vars)
+	env, err := NewEnv(r, vars)
 	if err != nil {
 		return EnvFile{}, err
 	}
@@ -388,8 +390,8 @@ func (e EnvFile) Secrets() []string {
 
 // NewEnv loads an environment of key-value pairs from a string.
 // The format of the string can be `key: value` or `key=value` pairs.
-func NewEnv(raw string, vars map[string]string) (EnvSource, error) {
-	env, parser, err := parseEnvironment(raw)
+func NewEnv(r io.Reader, vars map[string]string) (EnvSource, error) {
+	env, parser, err := parseEnvironment(r)
 	if err != nil {
 		return nil, err
 	}
@@ -432,13 +434,14 @@ type envvar struct {
 // It first tries the key=value format. When that returns an error,
 // the yml format is tried.
 // The default parser to be used with the format is also returned.
-func parseEnvironment(raw string) ([]envvar, tpl.Parser, error) {
+func parseEnvironment(r io.Reader) ([]envvar, tpl.Parser, error) {
 	parser := tpl.NewV2Parser()
-	env, err := parseEnv(raw)
+	var ymlReader bytes.Buffer
+	env, err := parseEnv(io.TeeReader(r, &ymlReader))
 	if err != nil {
 		var ymlErr error
 		parser = tpl.NewV1Parser()
-		env, ymlErr = parseYML(raw)
+		env, ymlErr = parseYML(&ymlReader)
 		if ymlErr != nil {
 			return nil, nil, err
 		}
@@ -446,9 +449,9 @@ func parseEnvironment(raw string) ([]envvar, tpl.Parser, error) {
 	return env, parser, nil
 }
 
-func parseEnv(raw string) ([]envvar, error) {
+func parseEnv(r io.Reader) ([]envvar, error) {
 	vars := map[string]envvar{}
-	scanner := bufio.NewScanner(strings.NewReader(raw))
+	scanner := bufio.NewScanner(r)
 
 	i := 1
 	for scanner.Scan() {
@@ -480,9 +483,14 @@ func parseEnv(raw string) ([]envvar, error) {
 	return res, nil
 }
 
-func parseYML(raw string) ([]envvar, error) {
+func parseYML(r io.Reader) ([]envvar, error) {
+	contents, err := ioutil.ReadAll(r)
+	if err != nil {
+		return nil, err
+	}
+
 	pairs := make(map[string]string)
-	err := yaml.Unmarshal([]byte(raw), pairs)
+	err = yaml.Unmarshal(contents, pairs)
 	if err != nil {
 		return nil, err
 	}
--- a/internals/secrethub/run.go
+++ b/internals/secrethub/run.go
@@ -130,13 +130,8 @@ func (cmd *RunCommand) Run() error {
 		}
 	}
 
-	client, err := cmd.newClient()
-	if err != nil {
-		return errio.Error(err)
-	}
-
 	if cmd.template != "" {
-		envFile, err := NewEnvFile(cmd.template, templateVars, newSecretReader(client))
+		envFile, err := NewEnvFile(cmd.template, templateVars)
 		if err != nil {
 			return err
 		}
@@ -161,6 +156,11 @@ func (cmd *RunCommand) Run() error {
 		}
 	}
 
+	client, err := cmd.newClient()
+	if err != nil {
+		return errio.Error(err)
+	}
+
 	for path := range secrets {
 		secret, err := client.Secrets().Versions().GetWithData(path)
 		if err != nil {
@@ -171,8 +171,9 @@ func (cmd *RunCommand) Run() error {
 
 	// Construct the environment, sourcing variables from the configured sources.
 	environment := make(map[string]string)
+	sr := newSecretReader(client)
 	for _, source := range envSources {
-		pairs, err := source.Env(secrets)
+		pairs, err := source.Env(secrets, sr)
 		if err != nil {
 			return errio.Error(err)
 		}
@@ -319,7 +320,7 @@ func parseKeyValueStringsToMap(values []string) (map[string]string, error) {
 // EnvSource defines a method of reading environment variables from a source.
 type EnvSource interface {
 	// Env returns a map of key value pairs.
-	Env(secrets map[string]string) (map[string]string, error)
+	Env(secrets map[string]string, sr tpl.SecretReader) (map[string]string, error)
 	// Secrets returns a list of paths to secrets that are used in the environment.
 	Secrets() []string
 }
@@ -327,15 +328,14 @@ type EnvSource interface {
 type envTemplate struct {
 	envVars      map[string]tpl.Template
 	templateVars map[string]string
-	secretReader tpl.SecretReader
 }
 
 // Env injects the given secrets in the environment values and returns
 // a map of the resulting environment.
-func (t envTemplate) Env(secrets map[string]string) (map[string]string, error) {
+func (t envTemplate) Env(secrets map[string]string, sr tpl.SecretReader) (map[string]string, error) {
 	result := make(map[string]string)
 	for key, template := range t.envVars {
-		value, err := template.Evaluate(t.templateVars, t.secretReader)
+		value, err := template.Evaluate(t.templateVars, sr)
 		if err != nil {
 			return nil, err
 		}
@@ -351,12 +351,12 @@ func (t envTemplate) Secrets() []string {
 }
 
 // NewEnvFile returns an new environment from a file.
-func NewEnvFile(filepath string, vars map[string]string, sr tpl.SecretReader) (EnvFile, error) {
+func NewEnvFile(filepath string, vars map[string]string) (EnvFile, error) {
 	content, err := ioutil.ReadFile(filepath)
 	if err != nil {
 		return EnvFile{}, ErrCannotReadFile(filepath, err)
 	}
-	env, err := NewEnv(string(content), vars, sr)
+	env, err := NewEnv(string(content), vars)
 	if err != nil {
 		return EnvFile{}, err
 	}
@@ -373,8 +373,8 @@ type EnvFile struct {
 }
 
 // Env returns a map of key value pairs read from the environment file.
-func (e EnvFile) Env(secrets map[string]string) (map[string]string, error) {
-	env, err := e.env.Env(secrets)
+func (e EnvFile) Env(secrets map[string]string, sr tpl.SecretReader) (map[string]string, error) {
+	env, err := e.env.Env(secrets, sr)
 	if err != nil {
 		return nil, ErrTemplateFile(e.path, err)
 	}
@@ -388,7 +388,7 @@ func (e EnvFile) Secrets() []string {
 
 // NewEnv loads an environment of key-value pairs from a string.
 // The format of the string can be `key: value` or `key=value` pairs.
-func NewEnv(raw string, vars map[string]string, sr tpl.SecretReader) (EnvSource, error) {
+func NewEnv(raw string, vars map[string]string) (EnvSource, error) {
 	env, parser, err := parseEnvironment(raw)
 	if err != nil {
 		return nil, err
@@ -411,7 +411,6 @@ func NewEnv(raw string, vars map[string]string, sr tpl.SecretReader) (EnvSource,
 	return envTemplate{
 		envVars:      secretTemplates,
 		templateVars: vars,
-		secretReader: sr,
 	}, nil
 }
 
@@ -536,7 +535,7 @@ func NewEnvDir(path string) (EnvDir, error) {
 }
 
 // Env returns a map of environment variables sourced from the directory.
-func (dir EnvDir) Env(secrets map[string]string) (map[string]string, error) {
+func (dir EnvDir) Env(secrets map[string]string, _ tpl.SecretReader) (map[string]string, error) {
 	return dir, nil
 }
 
@@ -567,7 +566,7 @@ func NewEnvFlags(flags map[string]string) (EnvFlags, error) {
 
 // Env returns a map of environment variables sourced from
 // command-line flags and set to their corresponding value.
-func (ef EnvFlags) Env(secrets map[string]string) (map[string]string, error) {
+func (ef EnvFlags) Env(secrets map[string]string, _ tpl.SecretReader) (map[string]string, error) {
 	result := make(map[string]string)
 	for name, path := range ef {
 		result[name] = secrets[path]
--- a/internals/secrethub/run.go
+++ b/internals/secrethub/run.go
@@ -131,7 +131,7 @@ func (cmd *RunCommand) Run() error {
 	}
 
 	if cmd.template != "" {
-		envFile, err := NewEnvFile(cmd.template, templateVars)
+		envFile, err := ReadEnvFile(cmd.template, templateVars)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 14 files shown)


internals/secrethub/tpl/v2.go
```diff
@@ -24,95 +37,418 @@ func NewV2Parser() Parser {
...
-// Secrets returns a list of paths to secrets that are used in the template.
-func (t secretTemplateV2) Secrets() []string {
-	set := map[string]struct{}{}
-	for _, path := range t.secrets {
-		set[path] = struct{}{}
+// parseSecret parses the contents of a secret tag up to the closing delimiter.
+// parseSecret should be called after the opening delimiter has been read. The next
+// character from the buffer should be the first character of the contents.
+//
+// when parseSecret returns, the next character in the buffer is the first character
```
mackenbach: Capital `W`

internals/secrethub/tpl/v2.go
```diff
@@ -24,95 +37,418 @@ func NewV2Parser() Parser {
...
+			return nil, err
+		}
+
+		switch p.current {
+		case '$':
+			switch p.next {
+			case '{':
+				err = p.readRune()
+				if err == io.EOF {
+					return nil, ErrVariableTagNotClosed(p.lineNo, p.columnNo+1)
```
mackenbach: Why do you do `columnNo+1` here?

internals/secrethub/tpl/v2.go
```diff
@@ -24,95 +37,418 @@ func NewV2Parser() Parser {
...
+			switch p.next {
+			case '}':
+				return secret{
+					path: path,
+				}, nil
+			default:
+				return nil, ErrIllegalSecretCharacter(p.current, p.lineNo, p.columnNo)
+			}
+		default:
+			if unicode.IsLetter(p.current) || unicode.IsDigit(p.current) || p.current == '_' || p.current == '-' || p.current == '.' || p.current == '/' || p.current == ':' {
```
mackenbach: Consider creating a small helper function for this, e.g. `func isValidSecretPathRune(r rune) bool` and while you're at it, create his buddy too: `isValidVariableRune`

internals/secrethub/tpl/v2.go
```diff
@@ -24,95 +37,418 @@ func NewV2Parser() Parser {
...
+			return nil, ErrIllegalSecretCharacter(p.current, p.lineNo, p.columnNo)
+		}
 	}
+}
 
-	result := make([]string, len(set))
+// forwardToClosing skips all spaces up to the closing delimiter.
+// It returns an error when characters other than spaces occur before the complete
+// closing delimiter occurs.
+func (p *v2Parser) forwardToClosing(delim []rune) error {
```
mackenbach: Call it `delimiters` or `delims` please to signal this can be multiple. Also, consider using the `...rune` instead so this is simple for both single and multiple delimieters:

```golang
parser.forwardToClosing('}')
parser.forwardToClosing('}', '}')
```

mackenbach: Also, consider renaming the function to something like `skipSpaceToClosingDelimiter` or just `skipTrailingSpace(delimiters ...rune)`

internals/secrethub/tpl/v2.go
```diff
@@ -24,95 +37,418 @@ func NewV2Parser() Parser {
...
+		if err != nil {
+			return err
+		}
+	}
 	i := 0
-	for path := range set {
-		result[i] = path
+	for {
+		if p.next != delim[i] {
+			return errors.New("expected end delimiter")
```
mackenbach: How useful is this error to the users? Expected what end delimiter?

internals/secrethub/tpl/v2_test.go
```diff
@@ -1,23 +1,469 @@
...
+						character('e'),
+						character('c'),
+						character('r'),
+						character('e'),
+						character('t'),
+					},
+				},
+			},
+		},
+		"{ and } chars used": {
```
mackenbach: How about adding a case with `{` and `}` characters wrapping a secret tag, e.g. `{{{path/to/secret}}}`?

internals/secrethub/tpl/v2_test.go
```diff
@@ -1,23 +1,469 @@
...
+				character('h'),
+				character(' '),
+				character('}'),
+				character('}'),
+			},
+		},
+		"escaped backslash": {
+			input: `\\`,
+			expected: []node{
+				character('\\'),
```
mackenbach: Haha this is confusing, because the `\` character needs to be escaped too in Golang. 😄

internals/secrethub/tpl/v2.go
SimonBarendse: This error never ends up at the user. It's converted to another error based on the context (e.g. invalid character, secret/variable tag not closed).

However, we could change the error a bit in case we'll use this function from somewhere else in the future that does not change the error message.

internals/secrethub/tpl/v2_test.go
SimonBarendse: 3196cc06cae09261e54eb0d39d23c21d1ed95f94

internals/secrethub/tpl/v2_test.go
SimonBarendse: Haha jup. With strings we can use backticks ``` ` ```, but as far as I know we don't have something similar for runes.

internals/secrethub/tpl/v2.go
SimonBarendse: 4238e0a189fcd4d7fe0d44ebb2df3e0aada1d559

SimonBarendse: The function also forwards to the end of the delimiter. Do you have a suggestion for the name that also reflects this?

internals/secrethub/tpl/v2.go
SimonBarendse: 137e5aac6ffba700841da3079422243304c266b1

SimonBarendse: 137e5aac6ffba700841da3079422243304c266b1

internals/secrethub/tpl/v2.go
```diff
@@ -24,95 +37,418 @@ func NewV2Parser() Parser {
...
+// parseVar parses the contents of a template variable up to the closing delimiter.
+// parseVar should be called after the opening delimiter has been read. The next
+// character from the buffer should be the first character of the contents.
+//
+// when parseVar returns, the next character in the buffer is the first character
+// after the closing delimiter of the template variable.
+func (p *v2Parser) parseVar() (node, error) {
+	var buffer bytes.Buffer
+
+	for p.next == ' ' {
```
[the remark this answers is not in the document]
SimonBarendse: I think we should allow tabs as well. I've changed this in 1f363184b128ed5b8ea8347addef7799d305aa4b.

internals/secrethub/tpl/v2.go
SimonBarendse: Replaced the function in 1f363184b128ed5b8ea8347addef7799d305aa4b

internals/secrethub/tpl/v2.go
SimonBarendse: e76bf2ff93dea5de39d1bd11fee1a631e4756589

internals/secrethub/tpl/v2.go
```diff
@@ -24,95 +37,418 @@ func NewV2Parser() Parser {
...
 
-// InjectVars takes a map of template variables with their corresponding values. It replaces
-// the template variables with their values in the template.
-func (t templateV2) InjectVars(vars map[string]string) (SecretTemplate, error) {
-	secrets := make(map[string]string, len(t.secrets))
-	for k, template := range t.secrets {
-		secretpath, err := template.Inject(vars)
+// parseVar parses the contents of a template variable up to the closing delimiter.
+// parseVar should be called after the opening delimiter has been read. The next
+// character from the buffer should be the first character of the contents.
```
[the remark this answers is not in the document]
SimonBarendse: 0e7a93cb0fb25138f34827a6f784fd92931520dd

internals/secrethub/tpl/v2.go
```diff
@@ -24,95 +37,418 @@ func NewV2Parser() Parser {
...
+	return err
+}
+
+func (p *v2Parser) parse() ([]node, error) {
+	res := []node{}
+	err := p.readRune()
+	if err == io.EOF {
+		return res, nil
+	}
+	if err != nil {
```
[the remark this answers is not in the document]
SimonBarendse: [Effective Go](https://golang.org/doc/effective_go.html#if) and golint have another opinion about that:
> In the Go libraries, you'll find that when an if statement doesn't flow into the next statement—that is, the body ends in break, continue, goto, or return—the unnecessary else is omitted.

internals/secrethub/run_test.go
```diff
@@ -41,44 +45,49 @@ func TestParseEnv(t *testing.T) {
...
-					lineNumber: 1,
+					key:          "key",
+					value:        "value",
+					lineNumber:   1,
+					columnNumber: 7,
 				},
 			},
 		},
 		"success with multiple spaces": {
 			raw: "key    = value",
```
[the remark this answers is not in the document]
SimonBarendse: > Consider adding a test case for key = value and checking your line number arithmetic.

Do you mean these https://github.com/secrethub/secrethub-cli/pull/103/files/f85f2e4ed925959c661bc5e3450a4da9bcdde170#diff-e4b7e205ba964ff3ae490bef6508122bR211 ?

> Also, PR #100 adds some test cases that you may want to cover as well.

Let's deal with those in #100

internals/secrethub/run.go
```diff
@@ -342,31 +344,19 @@ func (t envTemplate) Env(secrets map[string]string) (map[string]string, error) {
 	return result, nil
 }
 
-// Secrets returns a list of paths to secrets that are used in the environment.
+// Secrets implements the EnvSource.Secrets function.
+// The envTemplate fetches its secrets using a tpl.SecretReader.
 func (t envTemplate) Secrets() []string {
```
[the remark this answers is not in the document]
SimonBarendse: The `Secrets` function comes from the `EnvSource` interface. We can change the RunCommand later to not use that interface. In the meantime, `envTemplate` implements it.

internals/secrethub/run.go
```diff
@@ -398,33 +388,30 @@ func (e EnvFile) Secrets() []string {
 
 // NewEnv loads an environment of key-value pairs from a string.
 // The format of the string can be `key: value` or `key=value` pairs.
-func NewEnv(raw string, vars map[string]string) (EnvSource, error) {
+func NewEnv(raw string, vars map[string]string, sr tpl.SecretReader) (EnvSource, error) {
```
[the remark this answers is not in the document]
SimonBarendse: `NewEnv` creates a new env source. One part of that is parsing the raw string/file. Another constructing the resulting struct and adding the template variables to that struct. For the caller of the function, it's not important what exactly is done to construct the env source, just that it is done.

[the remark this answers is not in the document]
SimonBarendse: Good suggestion to let it accept a reader :+1:

internals/secrethub/run.go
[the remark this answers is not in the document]
mackenbach: Yes so your long term plans are to deprecate this, but short term we're keeping the interface?

internals/secrethub/run.go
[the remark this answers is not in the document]
SimonBarendse: I've changed `NewEnv` to accept a reader in 3944d860b0db5d30b6f5b9ef392eb5350e125c64

internals/secrethub/tpl/v2.go
```diff
@@ -266,49 +265,48 @@ func (p *v2Parser) parse() ([]node, error) {
...
-		case '}':
+		if p.next == '}' {
 			return variable{
 				key: buffer.String(),
 			}, nil
-		case ' ':
-			errIllegalVariableSpace := ErrIllegalVariableCharacter(p.next, p.lineNo, p.columnNo+1)
-			err := p.forwardToClosing('}')
+		}
+		if p.isAllowedWhiteSpace(p.next) {
```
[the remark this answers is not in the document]
mackenbach: Discussed and the `if` and `else if` thingy is apparently not idiomatic Go if each statement returns. You learn something new every day. 

Instead we discussed to add more logical grouping to the file with whitespace.

internals/secrethub/run_test.go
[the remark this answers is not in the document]
mackenbach: Ah I see what went wrong, the markdown formatter just stripped my extra whitespace. Meant to say with multiple spaces before `value`:

```
key =       value
```

I felt the multiple spaces thing is indeed unit tested by the syntax, but the correct calculation of the starting point of `value` may not be. That happens in this piece of code under test and could benefit from one edge case test. If you feel that has been tested already, then cool. 

> Let's deal with those in #100
👍

internals/secrethub/run.go
```diff
@@ -436,9 +423,10 @@ func templateError(envvar envvar, err error) error {
...
-	key        string
-	value      string
-	lineNumber int
+	key          string
+	value        string
+	lineNumber   int
+	columnNumber int
 }
 
 // parseEnvironment parses envvars from a string.
```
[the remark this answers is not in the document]
SimonBarendse: I've renamed `parseEnv` to `parseDotEnv`.

internals/secrethub/run_test.go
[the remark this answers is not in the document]
SimonBarendse: Ahh, yes, the formatter is a bit too eager to strip spaces in this case :sweat_smile: 

Good suggestion, I've added a testcase for multiple spaces before the value.

internals/secrethub/tpl/v2_test.go
```diff
@@ -390,6 +390,10 @@ func TestParserV2_parse(t *testing.T) {
 			input: "{{ a@b }}",
 			err:   ErrIllegalSecretCharacter('@', 1, 5),
 		},
+		"illegal { at start of secret tag": {
+			input: "{{{ path/to/secret }}}",
```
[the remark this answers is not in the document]
SimonBarendse: #105

internals/secrethub/tpl/internal/token/token.go
```diff
@@ -0,0 +1,42 @@
...
+	tokens = map[rune]struct{}{
+		Dollar:    {},
+		LBracket:  {},
+		RBracket:  {},
+		Backslash: {},
+	}
+)
+
+// IsDollar returns whether the given rune is the dollar token.
+func IsDollar(ch rune) bool {
```
jpcoenen: We discussed that these functions in itself don't really add any value because ` == ` is just as simple.

internals/secrethub/tpl/v2.go
```diff
@@ -24,95 +26,439 @@ func NewV2Parser() Parser {
...
+				continue
+			default:
+				// We don't allow dollars before letters and underscores now,
+				// as we might want to use these for $var support (without brackets) later.
+				if unicode.IsLetter(p.next) || p.next == '_' {
+					return nil, ErrUnexpectedDollar(p.lineNo, p.columnNo)
+				}
+
+				res = append(res, character(p.current))
+				continue
```
jpcoenen: I see a lot of these that are technically not necessary. I understand why you've added them: to force a break from this switch. So this made me think: can't this switch live in a function? That way, you can force the switch to end at the end of every case by returning. The compiler can then check that you haven't missed any returns (because you dont place a return at the end of the function, so every individual case must have a a return).

I can't easily see whether it is possible to extract this to a function without having to make the code ugly. Do you have any idea?

jpcoenen: Tried it and came up with this. What do you think?
```go
func (p *v2Parser) parse() ([]node, error) {
	res := []node{}

	err := p.readRune()
	if err == io.EOF {
		return res, nil
	}
	if err != nil {
		return nil, err
	}

	for {
		err := p.readRune()
		if err == io.EOF {
			return append(res, character(p.current)), nil
		}
		if err != nil {
			return nil, err
		}
		r, err := p.parseRune()
		if err == io.EOF {
			return append(res, r), nil
		}
		if err != nil {
			return nil, err
		}
		res = append(res, r)
	}
}

func (p *v2Parser) parseRune() (node, error) {
	switch p.current {
	case token.Dollar:
		switch p.next {
		case token.LBracket:
			variable, err := p.parseVar()
			if err != nil {
				return nil, err
			}
			return variable, p.readRune()
		default:
			// We don't allow dollars before letters and underscores now,
			// as we might want to use these for $var support (without brackets) later.
			if unicode.IsLetter(p.next) || p.next == '_' {
				return nil, ErrUnexpectedDollar(p.lineNo, p.columnNo)
			}
			return character(p.current), nil
		}
	case token.LBracket:
		switch p.next {
		case token.LBracket:
			secret, err := p.parseSecret()
			if err != nil {
				return nil, err
			}
			return secret, p.readRune()
		default:
			return character(p.current), nil
		}
	case token.Backslash:
		if token.IsToken(p.next) {
			r := character(p.next)
			return r, p.readRune()
		} else {
			return character(p.current), nil
		}
	default:
		return character(p.current), nil
	}
}
```

internals/secrethub/tpl/v2_test.go
```diff
@@ -1,23 +1,496 @@
...
 
-	"github.com/secrethub/secrethub-cli/internals/secrethub/tpl"
-	generictpl "github.com/secrethub/secrethub-cli/internals/tpl"
+	"github.com/secrethub/secrethub-cli/internals/secrethub/tpl/fakes"
+
 	"github.com/secrethub/secrethub-go/internals/assert"
 )
 
+func TestParserV2_parse(t *testing.T) {
+	cases := map[string]struct {
```
jpcoenen: Am I correct that the following case is missing? `{{ foo/bar }a` 

I couldn't get it to output a correct error. It gave "illegal character in secret ' '`. Or does that make sense?

internals/secrethub/tpl/v2.go
SimonBarendse: Agree that a `return` can be more intuitive and robust than `continue`s in a for loop.

Improved a bit further, as the `return character(p.current), nil` can be re-used.

SimonBarendse: 5eb4c0

SimonBarendse: See 5eb4c0b80cf9ace20e1032692fd5b19d9018135e

internals/secrethub/run_test.go
```diff
@@ -70,7 +70,7 @@ func TestParseDotEnv(t *testing.T) {
 				},
 			},
 		},
-		"success with multiple spaces": {
+		"success with multiple spaces before key": {
```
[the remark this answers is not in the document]
SimonBarendse: The case for multiple spaces before the key is added in #100

internals/secrethub/tpl/v2_test.go
SimonBarendse: After discussion we've chosen to rephrase the error to `unexpected character 'a', expected '}'`

internals/secrethub/tpl/v2.go
```diff
@@ -179,46 +179,32 @@ func (p *v2Parser) parse() ([]node, error) {
...
 func (p *v2Parser) parseRoot() (node, error) {
-	switch p.current {
-	case token.Dollar:
-		switch p.next {
-		case token.LBracket:
-			variable, err := p.parseVar()
-			if err != nil {
-				return nil, err
-			}
+	if p.current == token.Dollar && p.next == token.LBracket {
```
jpcoenen: Nice, much cleaner! Very readable :smile:

internals/secrethub/tpl/errors.go
```diff
@@ -31,6 +31,19 @@ func ErrUnexpectedDollar(lineNo, colNo int) error {
...
+// ErrUnexpectedCharacter is returned when expecting a specific character, for example
+// the first character of a closing delimiter after a space occurred in a tag, or
+// the second character of a closing delimiter after the first character of the closing
+// delimiter.
+func ErrUnexpectedCharacter(lineNo, colNo int, actual, expected rune) error {
+	return templateSyntaxError{
+		lineNo: lineNo,
+		colNo:  colNo,
+		code:   "unexpected character",
+		msg:    fmt.Sprintf("unexpected '%c' expected '%c'", actual, expected),
```
jpcoenen: I'd add a comma before `expected`.

internals/secrethub/tpl/v2.go
```diff
@@ -24,95 +26,392 @@ func NewV2Parser() Parser {
...
+		if err == io.EOF {
+			return ErrVariableTagNotClosed(p.lineNo, p.columnNo+1)
+		}
+		return err
+	}
+
+	err := checkError(p.readRune())
+	if err != nil {
+		return nil, err
+	}
```
mackenbach: This can be simplified to:

```
if err != nil {
   return checkError(err)
}
```

SimonBarendse: #108

PR #106, opened 2019-06-21
Add template syntax support for run env file keys

SimonBarendse:


internals/secrethub/run.go
```diff
@@ -404,7 +404,7 @@ func NewEnv(r io.Reader, vars map[string]string) (EnvSource, error) {
 			return nil, templateError(envvar, err)
 		}
 
-		template, err := parser.Parse(envvar.value, envvar.lineNumber, envvar.columnNumber)
+		template, err := parser.Parse(envvar.value, envvar.lineNo, envvar.valColNo)
```
jpcoenen: What's the reason for changing `Number` to `No`?

internals/secrethub/run.go
```diff
@@ -478,7 +479,16 @@ func parseDotEnv(r io.Reader) ([]envvar, error) {
 			valColNo++
 		}
 
+		keyColNo := 1 // one for the current column.
+		for _, r := range parts[0] {
+			if !unicode.IsSpace(r) {
+				break
+			}
+			keyColNo++
```
jpcoenen: What happens with a tab character?

internals/secrethub/run.go
```diff
@@ -339,12 +340,18 @@ type envvarTpls struct {
...
+	return "", ErrSecretsNotAllowedInKey
+}
+
 // Env injects the given secrets in the environment values and returns
 // a map of the resulting environment.
 func (t envTemplate) Env(secrets map[string]string, sr tpl.SecretReader) (map[string]string, error) {
 	result := make(map[string]string)
 	for _, tpls := range t.envVars {
-		key, err := tpls.key.Evaluate(t.templateVars, sr)
+		key, err := tpls.key.Evaluate(t.templateVars, secretReaderNotAllowed{})
```
jpcoenen: Seeing that we're not passing a pointer here, led me to looking at other usages of `Evaluate()`. Seems that we're always passing a value and in case of a normal `secretReader` that value contains the value of a `secrethub.Client`. Meaning that we're copying the client. Is that something we want?

SimonBarendse: Ahh, no, it's not. We want to have a pointer to a client, as the client can cache etc.

Actually, we probably want the SecretReader to accept a function that creates a client and have it call that function on the first call to ReadSecret. By doing so, we postpone creating the client, which postpones prompting the user for their passphrase.
For example, in case of a parse error, we don't want to prompt the user for their passphrase.
And for templates without any secret, we then won't prompt at all :+1:.

I'll change this in a new PR.

internals/secrethub/run.go
SimonBarendse: I didn't like the long name `columnNumberValue`. But `No` might be unclear.

I'll change it.

internals/secrethub/run.go
SimonBarendse:
```Go
unicode.IsSpace('\t')
// true
```

`unicode.IsSpace` checks for any whitespace, not just spaces.

jpcoenen: But is it okay that a tab is counted as only 1 column? My IDE counts it as 4.

internals/secrethub/run.go
```diff
@@ -454,11 +454,11 @@ func NewEnv(r io.Reader, vars map[string]string) (EnvSource, error) {
...
-	key      string
-	value    string
-	lineNo   int
-	keyColNo int
-	valColNo int
+	key               string
+	value             string
+	lineNumber        int
+	columnNumberKey   int
+	ColumnNumberValue int
```
jpcoenen: Accidental captical `C`?

SimonBarendse: yes, changed it.

internals/secrethub/run.go
SimonBarendse: Ahh, good question.

Does it always count it as 4 or do you also have file types for which it counts them as 2?

jpcoenen: It depends on the used tab-width. To be honest, I don't really know how to fix this nicely. Maybe we should just keep it at 1.

PR #107, opened 2019-06-24
Add template version flag to the run command

SimonBarendse:


internals/secrethub/run.go
```diff
@@ -133,8 +135,25 @@ func (cmd *RunCommand) Run() error {
...
 	}
 
+	var parser tpl.Parser
+	switch cmd.templateVersion {
+	case "auto":
+		// Leave it up to the env file parser to detect the template version to use;
+		// The latest version is used for .env formatted environment files, v1 is used
+		// for yml formatted environment files.
+		parser = nil
+	case "1":
```
jpcoenen: In the command reference we say `v1` is that what we want or `1`? I have a light preference for `v1`. Or do we want both?

internals/secrethub/run.go
```diff
@@ -391,12 +410,16 @@ func (e EnvFile) Secrets() []string {
 
 // NewEnv loads an environment of key-value pairs from a string.
 // The format of the string can be `key: value` or `key=value` pairs.
-func NewEnv(r io.Reader, vars map[string]string) (EnvSource, error) {
-	env, parser, err := parseEnvironment(r)
+func NewEnv(r io.Reader, vars map[string]string, parser tpl.Parser) (EnvSource, error) {
+	env, defaultParser, err := parseEnvironment(r)
```
jpcoenen: Is there a reason for evaluating this if `parser==nil`?

SimonBarendse: Yes, we still need `env` and `err` if `parser != nil`. If `parser == nil` we also need the `defaultParser`.

internals/secrethub/run.go
SimonBarendse: Let's accept both :+1:

internals/secrethub/inject.go
```diff
@@ -115,8 +115,12 @@ func (cmd *InjectCommand) Run() error {
 	switch cmd.templateVersion {
 	case "1":
 		parser = tpl.NewV1Parser()
+	case "v1":
```
jpcoenen: Did you know you can just do `case "v1", "1":`?

SimonBarendse: No, thnx :smile:

jpcoenen: You're welcome

PR #109, opened 2019-06-24
 Postpone creation of client in SecretReader 

SimonBarendse:


internals/secrethub/secret_reader.go
```diff
@@ -3,19 +3,39 @@ package secrethub
...
+	secret, err := client.Secrets().Versions().GetWithData(path)
 	if err != nil {
 		return "", err
 	}
+
 	return string(secret.Data), nil
 }
+
+func (sr secretReader) client() (secrethub.Client, error) {
+	if sr._client == nil {
```
jpcoenen: Should this functionality reside here or in a higher layer?

SimonBarendse: I've moved it to the client factory. That way, the client used in the secret reader of the run command is also used when fetching secrets for envvars supplied via a flag.

PR #110, opened 2019-06-24
Make template variables case insensitive

SimonBarendse:


SimonBarendse:
> Did you know you also have to strings.EqualFold() function to compare two strings case insensitively? (not that the current approach is not good; it's great. Just a fun-fact).

Note that by casting to lowercase before adding an entry to the map, a variable with other case overwrites an earlier entry. When we add two entries to the map with different case, we don't know which to use.

For example:
```bash
SECRETHUB_VAR_HELLO=there secrethub run -v hello=world
```

The `hello` from the flag should overwrite the `HELLO` from the environment variable.

So in this case, I don't think we can postpone a cast up to the point of comparison.

PR #111, opened 2019-06-24
 Add an in-file flag to inject 

### LINKED ISSUE #37: Add --template flag to inject command

SimonBarendse:
resolves #37

### DIFF
--- a/internals/secrethub/inject.go
+++ b/internals/secrethub/inject.go
@@ -63,8 +63,8 @@ func (cmd *InjectCommand) Register(r Registerer) {
 			units.HumanDuration(cmd.clearClipboardAfter),
 		),
 	).Short('c').BoolVar(&cmd.useClipboard)
-	clause.Flag("in-file", "The filename of a template file to inject.").StringVar(&cmd.inFile)
-	clause.Flag("out-file", "Write the injected template to a file instead of stdout.").StringVar(&cmd.outFile)
+	clause.Flag("in-file", "The filename of a template file to inject.").Short('i').StringVar(&cmd.inFile)
+	clause.Flag("out-file", "Write the injected template to a file instead of stdout.").Short('o').StringVar(&cmd.outFile)
 	clause.Flag("file", "").Hidden().StringVar(&cmd.outFile)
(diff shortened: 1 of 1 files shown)


internals/secrethub/inject.go
```diff
@@ -61,8 +63,10 @@ func (cmd *InjectCommand) Register(r Registerer) {
 			units.HumanDuration(cmd.clearClipboardAfter),
 		),
 	).Short('c').BoolVar(&cmd.useClipboard)
-	clause.Flag("file", "Write the injected template to a file instead of stdout.").StringVar(&cmd.file)
-	clause.Flag("file-mode", "Set filemode for the file if it does not yet exist. Defaults to 0600 (read and write for current user) and is ignored without the --file flag.").Default("0600").SetValue(&cmd.fileMode)
+	clause.Flag("in-file", "The filename of a template file to inject.").Short('i').StringVar(&cmd.inFile)
+	clause.Flag("out-file", "Write the injected template to a file instead of stdout.").Short('o').StringVar(&cmd.outFile)
+	clause.Flag("file", "").Hidden().StringVar(&cmd.outFile)
```
mackenbach: Could you add a comment to this line indicating it's aliased? If someone comes along in 6 months, it needs to be self-evident that this is an alias. 

```
// Alias for --out-file
```

SimonBarendse: We might want to create a `flag.Alias(name string)` function later to make it self-explanatory.

PR #113, opened 2019-06-24
 Add support for using template variables without brackets

SimonBarendse:


internals/secrethub/tpl/v2.go
```diff
@@ -188,8 +188,15 @@ func (p *v2Parser) parseRoot() (node, error) {
 		return variable, p.readRune()
 	}
 
-	if p.current == token.Dollar && (unicode.IsLetter(p.next) || p.next == '_') {
-		return nil, ErrUnexpectedDollar(p.lineNo, p.columnNo)
+	if p.current == token.Dollar && p.isVariableStartRune(p.next) {
+		variable, err := p.parseVarWithoutBrackets()
+		if err == io.EOF {
```
jpcoenen: Do we need this check here?

internals/secrethub/tpl/v2_test.go
```diff
@@ -91,6 +91,94 @@ func TestParserV2_parse(t *testing.T) {
 				},
 			},
 		},
+		"uppercase var without brackets": {
```
jpcoenen: What about a variable with an underscore?

internals/secrethub/tpl/v2.go
SimonBarendse: We could also just `return p.parseVarWithoutBrackets()`

jpcoenen: My thought exactly. Keeping it short; keeping it simple :smile:

PR #114, opened 2019-06-24
Release v0.24.0

SimonBarendse:
## Added
- Use `.env` syntax for run env files
- Add variable support to inject templates and run Environment Files
- Improved error messages for template parsing
- Use the `secrethub.env` run env file when it exists and no other file is specified
- Add an `--in-file` flag to the inject command.
- Add `--out-file` flag to the inject command (previously called `--file`)
- Add `--env-file` flag to the run command (previously called `--template`)
- Add `-o` and `-i` shorthands for specifying an output and/or input file in the inject command.

## Changed
- The inject command now uses the [v2 template syntax](https://secrethub.io/docs/reference/template-syntax/) (which has variable support) by default. [Migrate to the v2 syntax](https://secrethub.io/docs/reference/template-syntax/#migrating-from-v1-to-v2-syntax) or use the template version flag to get the old behavior: `--template-version v1`.

## Deprecated
- The `--file` flag of the inject command. Use `--out-file` instead.
- The `--template` flag of the run command. Use `--env-file` instead.

## Other
- Replace Travis CI with Circle CI


internals/secrethub/run.go
```diff
@@ -90,18 +101,65 @@ func (cmd *RunCommand) Run() error {
...
-		if err != nil {
-			return errio.Error(err)
+	osEnv, err := parseKeyValueStringsToMap(os.Environ())
+	if err != nil {
+		return errio.Error(err)
+	}
+
+	templateVars := make(map[string]string)
+
+	for k, v := range osEnv {
```
jpcoenen: Should we do this if `template_version=1`? It makes that the command does not run if any `SECRETHUB_VAR` variable is set.

internals/secrethub/run.go
```diff
@@ -174,7 +227,7 @@ func (cmd *RunCommand) Run() error {
 	maskStrings := make([][]byte, len(secrets))
 	i := 0
 	for _, val := range secrets {
-		maskStrings[i] = val
+		maskStrings[i] = []byte(val)
```
jpcoenen: This does not contain the values from `.env` files.

SimonBarendse: I've opened #115 to fix this.

internals/secrethub/run.go
jpcoenen: We discsussed this in person and concluded that this behaviour is undesirable, but because it is an edge case of a feature that is decprecated, we will leave it as is. Changing it requires too many changes.

PR #117, opened 2019-06-25
Release v0.24.1

SimonBarendse:
### Fixed
- Undo breaking change of `v0.24.0`: Auto detect the template syntax version to use instead of defaulting to the new template syntax.


internals/secrethub/run.go
```diff
@@ -137,21 +137,14 @@ func (cmd *RunCommand) Run() error {
...
-		parser = nil
-	case "1", "v1":
-		parser = tpl.NewV1Parser()
-	case "2", "v2":
-		parser = tpl.NewV2Parser()
-	case "latest":
-		parser = tpl.NewParser()
-	default:
-		return ErrUnknownTemplateVersion(cmd.templateVersion)
+	raw, err := ioutil.ReadFile(cmd.envFile)
```
mackenbach: Why are we reading a file now? What changed that we need this here?

mackenbach: It seems to me you're always reading the `.env` file. What happens when there's no `secrethub.env` file and no `--env-file` flag specified? And if you add `--envar` to that situation?

mackenbach:
Also, please update the PR text to reflect the release notes for this version. When doing that, consider adding/changing the original `v0.24.0` release notes.

PR #104, opened 2019-06-20
Updates for secrethub-go v0.21.0

jpcoenen:
Tests still have to be fixed.

PR #120, opened 2019-07-01
AWS Integration

jpcoenen:
Implements the AWS integration from https://github.com/secrethub/secrethub-go/pull/91 into the CLI.

### Changes
1. Update `secrethub-go` to `v0.21.0` (see https://github.com/secrethub/secrethub-cli/pull/104)
1. Introduce `secrethub service aws init` command to create services that use AWS for authentication and key decryption.
1. Add the `--use-aws` flag that uses AWS for authentication and key decryption.


internals/secrethub/service_aws_init.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/secrethub/secrethub-cli/internals/cli/ui"
+
+	"github.com/secrethub/secrethub-go/internals/api"
+	"github.com/secrethub/secrethub-go/internals/errio"
+)
+
+// Errors
+var (
+	ErrInvalidAWSRegion = errMain.Code("invalid_region").Error("invalid AWS region")
+	ErrRoleAlreadyTaken = errMain.Code("role_taken").Error("a service using that IAM role already exists")
```
mackenbach: Maybe explain the constraint here? E.g. `a service account with that IAM role already exists. Either assume the role to use SecretHub or create a service account with a different IAM role.`

internals/secrethub/service_aws_init.go
```diff
@@ -0,0 +1,95 @@
...
+	cfg := aws.NewConfig()
+	if cmd.region != "" {
+		_, ok := endpoints.AwsPartition().Regions()[cmd.region]
+		if !ok {
+			return ErrInvalidAWSRegion
+		}
+		cfg = cfg.WithRegion(cmd.region)
+	}
+
+	service, err := client.Services().AWS().Create(repo.Value(), cmd.description, cmd.kmsKeyID, cmd.role, cfg)
```
mackenbach: Why do we pass that config to the AWS call and not create a new client that is decorated with the AWS client, which allows us to make that function call without the 'one-time setup' arguments?

internals/secrethub/service_aws_init.go
```diff
@@ -0,0 +1,95 @@
...
+	if err == api.ErrCredentialAlreadyExists {
+		return ErrRoleAlreadyTaken
```
mackenbach: Not a big fan of this sentinel error check, but seeing as it doesn't dictate logic but merely adapts the error message I think it is okay to do here. What do you think?

internals/secrethub/service_aws_init.go
```diff
@@ -0,0 +1,95 @@
...
+			if delErr != nil {
+				fmt.Fprintf(cmd.io.Stdout(), "Failed to cleanup after creating an access rule for %s failed. Be sure to manually remove the created service account %s: %s\n", service.ServiceID, service.ServiceID, err)
+				return errio.Error(delErr)
+			}
+
+			return errio.Error(err)
+		}
+	}
+
+	fmt.Fprintf(cmd.io.Stdout(), "The service %s is now reachable through AWS when the role %s is assumed.\n", service.ServiceID, cmd.role)
```
mackenbach: Consider something like: `Successfully created a service account. Any host that assumes the IAM role %s can now automatically authenticate to SecretHub and fetch the secrets it has been given access to.`

internals/secrethub/service_aws_init.go
```diff
@@ -0,0 +1,95 @@
...
+
+	return nil
+}
+
+// Register registers the command, arguments and flags on the provided Registerer.
+func (cmd *ServiceAWSInitCommand) Register(r Registerer) {
+	clause := r.Command("init", "Create a new AWS service account attached to a repository.")
```
mackenbach: What is an AWS service? Let's try to make this self-explanatory, e.g. `Create a new service account that is tied to an AWS IAM role and given access to a repository.`

internals/secrethub/service_aws_init.go
```diff
@@ -0,0 +1,95 @@
...
+	clause.Flag("path", "The service account is attached to the repository in this path and when used together with --permission, an access rule is created on the directory in this path.").Required().SetValue(&cmd.path)
+	clause.Flag("kms-key-id", "ID of the KMS-key to be used for encrypting the service's account key.").Required().StringVar(&cmd.kmsKeyID)
+	clause.Flag("role", "ARN of the IAM role that should have access to this service account.").Required().StringVar(&cmd.role)
+	clause.Flag("region", "The AWS region that should be used").StringVar(&cmd.region)
```
mackenbach: What does it mean when a `region is used`? Does it mean only hosts within that region can call secrethub and assume this service's role?

internals/secrethub/service_aws_init.go
```diff
@@ -0,0 +1,95 @@
...
+	clause.Flag("desc", "A description for the service").StringVar(&cmd.description)
```
mackenbach: `A short description for the service account so you'll recognize it later.`

internals/secrethub/client_factory.go
```diff
@@ -23,24 +24,35 @@ func NewClientFactory(store CredentialStore) ClientFactory {
...
 	if f.client == nil {
-		credential, err := f.store.Get()
-		if err != nil {
-			return nil, err
+		if f.UseAWS {
+			client, err := secrethub.NewClientAWS(f.NewClientOptions())
+			if err != nil {
+				return nil, err
+			}
+
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

internals/secrethub/client_factory.go
```diff
@@ -23,24 +24,34 @@ func NewClientFactory(store CredentialStore) ClientFactory {
...
+			if err != nil {
+				return nil, err
+			}
+			f.client = client
+		} else {
+			credential, err := f.store.Get()
+			if err != nil {
+				return nil, err
+			}
+
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

internals/secrethub/account_init.go
```diff
@@ -289,7 +289,18 @@ func (cmd *AccountInitCommand) createAccountKey() error {
...
 
 	fmt.Fprint(cmd.io.Stdout(), "Finishing setup of your account...")
-	_, err = client.Accounts().Keys().Create()
+
+	credential, err := cmd.credentialStore.Get()
+	if err != nil {
+		return err
+	}
+
+	fingerprint, err := credential.Fingerprint()
```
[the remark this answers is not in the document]
jpcoenen: I am not really sure anymore. It made sense in some way to take the loading of the credential out of the client. But it could use a second look.

internals/secrethub/service_aws_init.go
jpcoenen: Somewhat the same for me. Not fond of it, but it works okay. However, I think it's not that difficult to change this server-side.

internals/secrethub/service_aws_init.go
jpcoenen: Mainly important for KMS. The KMS-key should be in this region.

mackenbach: Then let's say that in the description 😄

internals/secrethub/account_init.go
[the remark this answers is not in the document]
mackenbach: Seeing as we're already passing the credential to the `Keys().Create()` function, I think it might be better to do the `credential.Fingerprint()` call inside that function? Keeps this code a bit cleaner. The only reason I can think of to keep it this way is maybe if you wouldn't call `.Fingerprint()` for every use case?

jpcoenen:
@jpcoenen I'm up

internals/secrethub/service_ls_test.go
```diff
@@ -57,6 +66,62 @@ func TestServiceLsCommand_Run(t *testing.T) {
...
+			},
+			serviceService: fakeclient.ServiceService{
+				Lister: fakeclient.RepoServiceLister{
+					ReturnsServices: []*api.Service{
+						{
+							ServiceID:   "test",
+							Description: "foobar",
+							Credential: &api.Credential{
+								Type: api.CredentialTypeAWSSTS,
+								Metadata: map[api.CredentialMetadataKey]string{
```
golangcibot: CredentialMetadataKey not declared by package api (from `typecheck`)

internals/secrethub/service_aws_init.go
```diff
@@ -56,6 +56,10 @@ func (cmd *ServiceAWSInitCommand) Run() error {
 		return err
 	}
 
+	if cmd.role == "" && cmd.kmsKeyID == "" {
+		fmt.Fprintln(cmd.io.Stdout(), "This command creates a new service account for use on AWS. For help on this, see https://secrethub.io/docs/reference/service-command/#command-service-aws-init.")
```
mackenbach: This is a dead link at the moment. Also, shouldn't you say that certain flags are missing?

internals/cli/ui/ask_test.go
```diff
@@ -244,15 +244,36 @@ func TestChoose(t *testing.T) {
 			in: []string{"\n", "2\n"},
 
 			expected: "bar",
-			out:      "foo? Press [ENTER] for options.\nfoo?\n1) foo\n2) bar\n3) baz\n",
+			out: "foo? Press [ENTER] for options.\n" +
+				"1) foo\n" +
+				"2) bar\n" +
+				"3) baz\n" +
+				"Give number of choice or enter other option: \n",
```
mackenbach: Two things that aren't immediately clear about the current wording:

- `Give number` is ambiguous, `type a number` is more direct.
- `enter other option` is ambiguous: the options are numbered so I can enter a different number than on the list or I can enter a different answer to the question (in our case a `region`)? `or enter a different %s` can be clearer. 

Instead of saying `Give number of choice or enter other option:` could we say `Type a number of choice or enter a different [region|foo|option]:`

internals/cli/ui/ask_test.go
```diff
@@ -244,15 +244,36 @@ func TestChoose(t *testing.T) {
...
 
 			expected: "Option 7",
-			out:      "foo? Press [ENTER] for options.\nfoo?\n1) Option 1\n2) Option 2\n3) Option 3\n4) Option 4\n5) Option 5\nPress [ENTER] for more options.\nfoo?\n1) Option 1\n2) Option 2\n3) Option 3\n4) Option 4\n5) Option 5\n6) Option 6\n7) Option 7\n8) Option 8\n9) Option 9\n10) Option 10\nPress [ENTER] for more options.\n",
+			out: "foo? Press [ENTER] for options.\n" +
+				"1) Option 1\n" +
+				"2) Option 2\n" +
+				"3) Option 3\n" +
+				"4) Option 4\n" +
+				"5) Option 5\n" +
+				"Give number of choice or enter other option ([ENTER] for more options): \n" +
```
mackenbach: This message is confusing to me. Either I type a number, enter a different region (/answer to foo question) and press ENTER, or I provide an 'empty' answer and just press ENTER to see more options. 

Can we make these use cases a bit more intuitive?

internals/secrethub/service_aws_init.go
```diff
@@ -81,8 +146,8 @@ func (cmd *ServiceAWSInitCommand) Run() error {
 func (cmd *ServiceAWSInitCommand) Register(r Registerer) {
 	clause := r.Command("init", "Create a new AWS service account attached to a repository.")
 	clause.Arg("repo", "The service account is attached to the repository in this path.").Required().SetValue(&cmd.repo)
-	clause.Flag("kms-key-id", "ID of the KMS-key to be used for encrypting the service's account key.").Required().StringVar(&cmd.kmsKeyID)
-	clause.Flag("role", "ARN of the IAM role that should have access to this service account.").Required().StringVar(&cmd.role)
+	clause.Flag("kms-key-id", "ID of the KMS-key to be used for encrypting the service's account key.").StringVar(&cmd.kmsKeyID)
```
mackenbach: It looks like we start sentences for other args/flags with `The XXX ...`. Consider adding that here and the line below too to be consistent.

internals/cli/ui/ask_test.go
SimonBarendse: > the options are numbered so I can enter a different number than on the list or I can enter a different answer to the question (in our case a region)?

Haha, it's intended as being a number from the list or a value (can be from the list).

e.g.:
1) foo
2) bar

Inputs can be: `1`, `2`, `foo` and `bar`, but also `baz`.

In the case of regions, only values from the list are valid, as we show all regions.

So I would omit `different` from the suggested phrase, to put let emphasis on that. How do you like that?

`Type a number of choice or enter a [region|foo|option]:`

mackenbach: Nice, maybe `Type a number of choice or enter a value` is clear enough, but I like it also with `region` in there.

internals/secrethub/service_aws_init.go
SimonBarendse: > Also, shouldn't you say that certain flags are missing?

You will be prompted for these values later. As users should be familiar with the concept of AWS services in order to know what values to supply, we thought the help should be referenced for users that have not created an AWS service before.
When the flags are supplied, you probably know about the concept, as you know what values you want to use.

SimonBarendse: > This is a dead link at the moment.

Should we reference a link or just add a longer help-text to `secrethub service aws init` (with long help)?

internals/secrethub/service_aws_init.go
SimonBarendse: Yes, let's add something about the IAM role :+1: 

What does `given access to a repository` mean? I think that might be confusing, as access rules still have to be set.

SimonBarendse: If we change that, let's change it for `secrethub service init` as well, as we also use `attach` there.

internals/secrethub/service_aws_init.go
SimonBarendse: Changed the help text to `The AWS region that should be used for KMS.`

internals/secrethub/service_aws_init.go
SimonBarendse: Renamed to `Create a new service account that is tied to an AWS IAM role` (being attached to a repo is explained in the repo arg help text and isn't central to the command).

internals/secrethub/service_aws_init.go
SimonBarendse: :+1: 

I've replaced the last `it` with `the service`, as I read it as referring to the host.

`Successfully created a service account. Any host that assumes the IAM role %s can now automatically authenticate to SecretHub and fetch the secrets the service has been given access to.`

internals/secrethub/account_init.go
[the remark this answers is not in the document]
jpcoenen: Added `Verifier` as an argument. Fingerprint can now be retrieved by `Create()`

internals/cli/ui/ask_test.go
jpcoenen: What about:
```
Give the number of an option or type a region ([ENTER] for more options):
```

internals/secrethub/client_factory.go
```diff
@@ -23,23 +24,29 @@ func NewClientFactory(store CredentialStore) ClientFactory {
...
 type clientFactory struct {
 	ServerURL *url.URL
+	UseAWS    bool
 	store     CredentialStore
 }
 
 // Register the flags for configuration on a cli application.
 func (f *clientFactory) Register(r FlagRegisterer) {
 	r.Flag("api-remote", "The SecretHub API address, don't set this unless you know what you're doing.").Hidden().URLVar(&f.ServerURL)
+	r.Flag("use-aws", "Use AWS credentials for authentication and account key decryption").BoolVar(&f.UseAWS)
```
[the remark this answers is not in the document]
SimonBarendse: Added the `--identity-provider` flag.

internals/secrethub/account_init.go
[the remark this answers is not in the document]
SimonBarendse: We decided not to pick this up right now.

internals/secrethub/service_aws_init.go
jpcoenen: A suggestion for a long help text:
```
The native AWS identity provider uses a combination of AWS IAM and AWS KMS to provide access to SecretHub for any service running on AWS (e.g. EC2, Lambda or ECS). For this to work, an IAM role and KMS key are needed. The role should be the role that is assumed by the service during executing: the "IAM Role" for an EC2 instance, "Task Execution Role" for an Lambda Function and "Task Role" for ECS Tasks.

The KMS key is a key that is used for encrypting the account key of the service. Decryption permission on this key must be granted to the previously described IAM Role.

To create a new service that uses the AWS identity provider, the CLI must have encryption access to the KMS key that will be used by the service account. Therefore AWS credentials should be configured on this system. For details on how this can be done, see https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html.

If no system-wide default for the AWS region is provided (e.g. with $AWS_REGION), the AWS-region where the KMS key resides should be explicitly provided to this command.
```

mackenbach: Amazing! Very concise and clear. I made a few grammatical and style fixes to your text. (shown in bold)

> The native AWS identity provider uses a combination of AWS IAM and AWS KMS to provide access to SecretHub for any service running on AWS (e.g. EC2, Lambda or ECS). For this to work, an IAM role and **a KMS** key are needed. 
>
> - The role should be the role that is assumed by the service during **execution**: the "IAM Role" for an EC2 instance, "Task Execution Role" for an Lambda Function and "Task Role" for ECS Tasks.
> - The KMS key is a key that is used for encrypting the account key of the service. Decryption permission on this key must be granted to the previously described IAM Role.
>
> To create a new service that uses the AWS identity provider, the CLI must have encryption access to the KMS key that will be used by the service account. Therefore AWS credentials should be configured on this system. For details on how this can be done, see https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html.
>
> If no system-wide default for the AWS region is provided (e.g. with $AWS_REGION), the AWS-region where the KMS key resides should be explicitly provided to this command **with the --region flag.**

Small consideration, we can make the second bullet even shorter by saying: `The KMS key is a key that is used for encryption of the account.` Not necessary, but worth thinking over for a sec.

jpcoenen: Good changes. I think it's worth mentioning that the role should have access. This can prevent a nasty headache.

PR #123, opened 2019-07-03
 Fix races in masked writer

SimonBarendse:


internals/cli/masker/writer.go
```diff
@@ -155,49 +166,48 @@ func (mw *MaskedWriter) flushBuffer() {
...
-					matcher.Reset()
-				}
-				mw.flushBuffer()
-			}
-			mw.lock.Unlock()
 		case output := <-mw.output:
 			for _, b := range output {
 				var err error
 				if b.masked {
 					if !masking {
```
mackenbach: This can definitely use a bit of documentation. If you're not already started masking the incoming bytes from `output` channel, write the full mask string `<redacted by SecretHub>` and keep dropping the bytes marked for masking until you reach the end of the byte sequence to mask.

internals/cli/masker/writer.go
```diff
@@ -85,61 +85,72 @@ type MaskedWriter struct {
...
+	buf              []maskByte
+	incomingBytes    chan []byte
+	forceFlushBuffer chan struct{}
+	output           chan []maskByte
+	err              chan error
+	wg               sync.WaitGroup
 }
 
 // NewMaskedWriter returns a new MaskedWriter that masks all occurrences of sequences in masks with maskString.
 func NewMaskedWriter(w io.Writer, masks [][]byte, maskString string, timeout time.Duration) *MaskedWriter {
```
mackenbach: I'd expect `maskString` to be called `mask` and `masks` to be called `sequencesToMask` or `bytesToMask`.

internals/cli/masker/writer.go
```diff
@@ -85,61 +85,72 @@ type MaskedWriter struct {
...
+			for _, b := range p {
+				mw.buf = append(mw.buf, maskByte{byte: b})
 
-		if !matchInProgress {
-			mw.flushBuffer()
-		}
+				for _, matcher := range mw.matchers {
+					maskLen := matcher.Read(b)
+					for i := 0; i < maskLen; i++ {
+						mw.buf[len(mw.buf)-1-i].masked = true
```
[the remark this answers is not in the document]
mackenbach: Although, if you change the names a bit that might be enough. Hard to say now though.

internals/cli/masker/writer.go
```diff
@@ -155,49 +166,48 @@ func (mw *MaskedWriter) flushBuffer() {
...
 					if err != nil {
-						mw.err = err
+						mw.err <- err
 						return
 					}
 					masking = false
 				}
 			}
-			mw.nOut += int64(len(output))
+			mw.wg.Add(-len(output))
```
mackenbach: Add a comment to this please (and maybe some whitespace). It is one of the crucial control flows in this entire file. There's only three places where it's used, all in different places: once for adding, once for subtracting, and once for waiting. It's nice if you're able to spot this at a glance.

internals/cli/masker/writer.go
```diff
@@ -155,49 +166,48 @@ func (mw *MaskedWriter) flushBuffer() {
...
+			select {
+			case mw.forceFlushBuffer <- struct{}{}:
+			default:
+			}
 		}
 	}
 }
 
 // Flush is called to make sure that all output is written to the underlying io.Writer.
 // Returns any errors caused by the writing.
```
mackenbach: Does it return any errors or does it return the first error that occurred during writing? 

```go
// Flush ensures that all output is written to the underlying io.Writer 
// and returns an error when a MaskedWriter.Write call fails.
```

Also, what is the effect of calling flush on the writer? Does it then no longer accept Write calls or will it keep working? If so, let's be clear about that in the comment. 

Finally, just by looking at the code, I'd expect the `Flush` function to call e.g. `forceFlush` but instead it just waits until the `MaskedWriter` is done processing. It actually doesn't flush a buffer to the output, that is already done by the run logic itself. 

Take a look at e.g. the [tabwriter](https://golang.org/pkg/text/tabwriter/). That too implements an `io.Writer` and has a `Flush` function. Instead of continuously passing on the data, it buffers it and adds tabs where necessary until `Flush` is called and the buffer is written to the underlying writer. With that architecture in mind, how do you see the architecture of the `MaskedWriter` and could we want to change that? E.g. should we consider making the `MaskedWriter` buffered with a real `Flush` function and separating the `Run` logic from it in e.g. a helper function or type? Something like `masker.Stream` with functions `Stream.Start` (maps to `MaskedWriter.Run`) and `Stream.Stop`(maps to `MaskedWriter.Flush`)?

internals/cli/masker/writer.go
```diff
@@ -155,49 +166,48 @@ func (mw *MaskedWriter) flushBuffer() {
...
 					}
 					masking = false
 				}
 			}
-			mw.nOut += int64(len(output))
+			mw.wg.Add(-len(output))
+		case <-time.After(mw.timeout):
+			// force the buffer to flush if not already done so.
+			select {
+			case mw.forceFlushBuffer <- struct{}{}:
```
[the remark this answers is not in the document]
mackenbach: Or don't if it should only ever be called in that place.

internals/cli/masker/writer.go
```diff
@@ -155,49 +166,48 @@ func (mw *MaskedWriter) flushBuffer() {
 //
 // This should be run in a separate goroutine.
```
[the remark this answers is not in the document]
SimonBarendse: The function should not start a goroutine. See [this discussion](https://github.com/secrethub/secrethub-cli/pull/78/files#r285085082) we've had about that. We could append `as it is blocking` to the comment to make clear why it should be run in a separate goroutine.

SimonBarendse:
> is this a fix or a feature?

This is a bugfix. We plan to release it together with the two bugfixes that are already merged to `develop` (#119, #122) and with #124.

That said, I've processed your feedback related to this bugfix. You've suggested some other changes that improve maintainability, which I'd like to pick up in a separate PR that can go in the next minor release.

internals/cli/masker/writer.go
```diff
@@ -195,7 +196,7 @@ func (mw *MaskedWriter) Run() {
 		case <-time.After(mw.timeout):
 			// force the buffer to flush if not already done so.
```
mackenbach: This comment still refers to a buffer flush. Is that still correct after your renaming to `outputTimeout`?

internals/cli/masker/writer.go
```diff
@@ -122,7 +122,16 @@ func (mw *MaskedWriter) Write(p []byte) (n int, err error) {
 	return len(p), nil
 }
 
-func (mw *MaskedWriter) write() {
+// process receives incoming bytes from calls to Write (over the incoming bytes channel) masks them if necessary and passes
+// them through to Run (over the output channel).
+//
+// - When the incoming bytes do not end in a partial secret, they are directly passed to the output channel.
+// - When the incoming bytes do end in a partial secret, all bytes up to the partial secret are passed to the
```
mackenbach: What happens when the incoming bytes **start** with a partial secret? Your phrasing here suggests that is a different scenario while I feel they may actually be the same: when the incoming bytes match a partial secret?

mackenbach:
Discussed and decided to push the bugfix through and pick it up later.

internals/cli/masker/writer.go
SimonBarendse: When incoming bytes start with a partial secret and other bytes that do not finish the secret come after, we know that the partial secret is not the secret, so we will output the plaintext of the partial secret.

For example:
There's one secret with the value `foo`

When the value `fbar` is written, all bytes are passed though (and outputted by the `Run` process), including the partial secret `f`.
When the value `barf` is written, `bar` is written, but the partial secret `f` is not yet. It is processed as described in the comment.

Does that clarify things? Do you feel the comment should be adapted to make this clearer?

PR #124, opened 2019-07-03
Warn about lowering the run masking-timeout flag value in the help

SimonBarendse:


internals/secrethub/run.go
```diff
@@ -82,7 +82,7 @@ func (cmd *RunCommand) Register(r Registerer) {
 	clause.Flag("var", "Define the value for a template variable with `VAR=VALUE`, e.g. --var env=prod").Short('v').StringMapVar(&cmd.templateVars)
 	clause.Flag("env", "The name of the environment prepared by the set command (default is `default`)").Default("default").Hidden().StringVar(&cmd.env)
 	clause.Flag("no-masking", "Disable masking of secrets on stdout and stderr").BoolVar(&cmd.noMasking)
-	clause.Flag("masking-timeout", "The time to wait for a partial secret that is written to stdout or stderr to be completed for masking.").Default("1s").DurationVar(&cmd.maskingTimeout)
+	clause.Flag("masking-timeout", "The time to wait for a partial secret that is written to stdout or stderr to be completed for masking. When the time between two partial writes of a secret is larger than this timeout, the secret is not masked. Warning: lowering this value increases the chance of secrets not being masked.").Default("1s").DurationVar(&cmd.maskingTimeout)
```
mackenbach: This description contains quite a bit of knowledge the user doesn't have about the inner workings of the masker. Let's try to word this in terms the user understands. How about something like this?

```
The masker listens for bytes written to stdout and stderr and this 
timeout value allows you to tweak the maximum time the masker 
will wait for the next bytes after it has detected a partial secret in 
those output streams. When the time between two partial writes 
of a secret is larger than this timeout, the secret is not masked. 
Warning: lowering this value reduces the odds of successfully 
masking a secret when it is written with long pauzes in between.
``` 

What I am still missing in this flag description is why it would be useful to the user to actually configure this number. It now talks about the risk of lowering it (and rightfully so), but doesn't give me an idea as to why I would ever use this flag. It hints at it with the words timeout and wait period, but it doesn't mention the (very small) latency that run introduces on stderr and stdout. 

So if you're satisfied with this flag description as I wrote above it's fine by me, but it does make me wonder (again) why we want to expose the configuration option in the first place.

SimonBarendse: > What I am still missing in this flag description is why it would be useful to the user to actually configure this number.

There's a trade-off between latency on stdout/stderr and security here. Each application can balance these two differently. For an application where latency of writes to stdin/stdout is super important, it might make sense to choose a lower timeout (e.g. 0). We've chosen a *default* that is more secure (1s), as that is more important to *most applications* than a 1s latency on stdout/stderr. By creating a flag to configure this value, developers of applications can make their own trade-off between latency on stdout/stderr and security. Because we've chosen a sensible default, they don't have to.

Should we make extra clear that secret masking is meant as an extra layer of security? i.e. that you still should not let your application write secrets to stdout/stderr and depend on secret masking to mask them.
That makes me thinking that we should probably also log a warning when such an event occurs, to notify developers, so that they know about it and can fix it.

Masking being an *extra layer* of security could also be important to know when making the trade-off, so that you know exactly what the risk is.

SimonBarendse: This intended use of masking should probably be explained in the general help text of the command, as it is not specific to the `--masking-timeout` flag.

SimonBarendse: That said, we could add another feature of still monitoring for partial secrets to be finished after the timeout and trigger a warning that the secret is leaked if the rest of the secret comes in later (because we then know this was indeed the secret). As again, developers want to know about this.

mackenbach: > This intended use of masking should probably be explained in the general help text of the command, as it is not specific to the --masking-timeout flag.

Yes agreed, put it wherever it makes the most sense. The lack of explanation on how masking works in the general help text means we have to explain a lot more in the flag documentation itself. If we explain how masking works and that it introduces a (very) small latency and to use `--mask-timeout` to configure it and how that is a tradeoff between latency and security I think we can be a lot more to the point in the flag description.

> Should we make extra clear that secret masking is meant as an extra layer of security? i.e. that you still should not let your application write secrets to stdout/stderr and depend on secret masking to mask them.

Yes definitely. Make it abundantly clear that this is to protect against bad behavior and that it's a best effort to protect, not the ultimate guarantee.

internals/secrethub/run.go
```diff
@@ -74,15 +74,22 @@ func NewRunCommand(newClient newClientFunc) *RunCommand {
 
 // Register registers the command, arguments and flags on the provided Registerer.
 func (cmd *RunCommand) Register(r Registerer) {
-	clause := r.Command("run", "Pass secrets as environment variables to a process.")
+	const helpShort = "Pass secrets as environment variables to a process."
+	const helpLong = `pass secrets as environment variables to a process.
+
+To protect against secrets leaking via stdout and stderr, those output streams are monitored for secrets. Detected secrets are masked by replacing them with `+ maskString + `.
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
To protect against secrets leaking via stdout and stderr, those output streams are monitored for secrets. Detected secrets are masked by replacing them with ` + maskString + `.
```

PR #128, opened 2019-07-05
Add service ls command

SimonBarendse:


internals/secrethub/service.go
```diff
@@ -21,4 +21,5 @@ func (cmd *ServiceCommand) Register(r Registerer) {
 	clause := r.Command("service", "Manage service accounts.")
 	NewServiceDeployCommand(cmd.io).Register(clause)
```
mackenbach: Not something your PR changes, but let's order the commands logically: `init, ls, deploy`.

mackenbach: Looked into it and it isn't a clear cut solution to switch the order around. We'll pick this up later.

PR #133, opened 2019-08-12
Add flag to the run command to suppress errors of non-existing secrets

### LINKED ISSUE #131: Add a --silent flag to the read / run command
There are many scenarios where some secrets could be missing at a given time and I would like to silently ignore it so that the cli does not emit the following message:
`Encountered an error: Secret not found (server.secret_not_found)`

Description:
>--silent
Silent or quiet mode. Don't show error messages.

SimonBarendse:
Resolves #131

This PR includes a fix for handling empty secret values, as secret masking could not handle those yet.

### DIFF
--- a/internals/secrethub/run.go
+++ b/internals/secrethub/run.go
@@ -228,7 +228,7 @@ func (cmd *RunCommand) Run() error {
 		cmd.command = strings.Split(cmd.command[0], " ")
 	}
 
-	secretsRead := secretReader.SecretsRead()
+	secretsRead := secretReader.Values()
 
 	maskStrings := make([][]byte, 0, len(secretsRead))
 	i := 0
--- a/internals/secrethub/run.go
+++ b/internals/secrethub/run.go
@@ -228,15 +228,19 @@ func (cmd *RunCommand) Run() error {
 
 	secretsRead := secretReader.SecretsRead()
 
-	maskStrings := make([][]byte, len(secrets)+len(secretsRead))
+	maskStrings := make([][]byte, 0, len(secrets)+len(secretsRead))
 	i := 0
 	for _, val := range secrets {
-		maskStrings[i] = []byte(val)
-		i++
+		if val != "" {
+			maskStrings[i] = []byte(val)
+			i++
+		}
 	}
 	for _, val := range secretsRead {
-		maskStrings[i] = []byte(val)
-		i++
+		if val != "" {
+			maskStrings[i] = []byte(val)
+			i++
+		}
 	}
 
 	maskedStdout := masker.NewMaskedWriter(os.Stdout, maskStrings, maskString, cmd.maskingTimeout)
--- a/internals/secrethub/run.go
+++ b/internals/secrethub/run.go
@@ -181,19 +181,15 @@ func (cmd *RunCommand) Run() error {
 		}
 	}
 
-	for path := range secrets {
-		client, err := cmd.newClient()
-		if err != nil {
-			return err
-		}
-		secret, err := client.Secrets().Versions().GetWithData(path)
-		if err != nil {
-			return err
-		}
-		secrets[path] = string(secret.Data)
-	}
-
 	secretReader := newBufferedSecretReader(newSecretReader(cmd.newClient))
+
+	for path := range secrets {
+		secret, err := secretReader.ReadSecret(path)
+		if err != nil {
+			return err
+		}
+		secrets[path] = secret
+	}
 
 	// Construct the environment, sourcing variables from the configured sources.
 	environment := make(map[string]string)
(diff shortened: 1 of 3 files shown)


internals/secrethub/run_test.go
```diff
@@ -454,6 +457,48 @@ func TestRunCommand_Run(t *testing.T) {
 				command: []string{"echo", "test"},
 			},
 		},
+		"missing secret": {
+			command: RunCommand{
+				command: []string{"echo", "test"},
+				envar:                map[string]string{
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
				envar: map[string]string{
```

internals/secrethub/secret_reader.go
```diff
@@ -58,3 +61,23 @@ func (sr *bufferedSecretReader) ReadSecret(path string) (string, error) {
...
+	return &ignoreMissingSecretReader{
+		secretReader: sr,
+	}
+}
+
+// ReadSecret uses the underlying secret reader to read the secret, but ignores
+// errors for non-existing secrets. Instead, it returns the empty string.
+func (sr *ignoreMissingSecretReader) ReadSecret(path string) (string, error) {
+	secret, err := sr.secretReader.ReadSecret(path)
+	if err == api.ErrSecretNotFound {
```
mackenbach: Question: what happens when say the repo doesn't exist? Does the server return `api.ErrSecretNotFound`? And by extension, how does this impact our ability to correctly ignore missing secrets? E.g. do we want to still ignore this when a different error (such as `api.ErrRepoNotFound`) is returned?

Consider whether we **can and want to** check the status code `404` instead of the exact error type to reduce coupling between the API implementation and the CLI.

internals/secrethub/run_test.go
```diff
@@ -454,6 +457,48 @@ func TestRunCommand_Run(t *testing.T) {
...
+						SecretService: &fakeclient.SecretService{
+							VersionService: &fakeclient.SecretVersionService{
+								WithDataGetter: fakeclient.WithDataGetter{
+									Err: api.ErrSecretNotFound,
+								},
+							},
+						},
+					}, nil
+				},
+				ignoreMissingSecrets: true,
```
mackenbach: What does this test? Where do we check that the resulting environment contains `MISSING=""`?

internals/secrethub/run.go
```diff
@@ -228,15 +228,19 @@ func (cmd *RunCommand) Run() error {
 
 	secretsRead := secretReader.SecretsRead()
```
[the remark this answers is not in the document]
SimonBarendse: It returns a list of all values that are read with the reader (so the returned secret values, not the given secret paths).

I think the ambiguity comes from the present and past tense of the verb read being the same.

Do you prefer the `ListSecrets` name? I feel like this doesn't cover it completely either, but I'm having trouble coming up with a better name.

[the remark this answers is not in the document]
SimonBarendse: How do you feel about `SecretsUsed`? It's not perfect either (as the secrets are not being used by the reader, so this is somewhat out of scope), but it makes a bit clearer that this is about the history of the reader.

Maybe `SecretsReturned` is better, as this limits to the scope of the reader.

internals/secrethub/run_test.go
SimonBarendse: It tests that no error is returned. It does indeed not test that the environment of the run command contains the missing key with the empty string as value.

One way to unit-test this would be to extract the logic of constructing the environment from the command, so that it can be tested in isolation (no need to inspect the environment of the other goroutine running the command). This should be done for non-empty secrets as well.

The e2e case could for example be tested by injecting a buffer as stdout and print the secrets to that buffer using the `echo` command (and use no-masking to show them).

In general, this command could use some more testing, both unit- and e2e tests.

internals/secrethub/secret_reader.go
SimonBarendse: You are right. The server indeed returns a different error when the repo does not exists.
I'd say this error should also be ignored when the flag is used.

As for whether this is possible: I've opened https://github.com/secrethub/secrethub-go/pull/96 to add functionality to support this.

internals/secrethub/run_test.go
mackenbach: Alright, if you feel it's tested thoroughly enough for now I'm cool with it. We can schedule more extensive testing for later if you want or add it now. Your call.

internals/secrethub/secret_reader.go
mackenbach: Cool, approved that PR. You decide wether to hold on merging this PR or change this later.

SimonBarendse: I've added the helper function in this PR, so that it isn't blocked by a release of secrethub-go. We can update it later to use the function of secrethub-go, once that is released.

Issue #140, opened 2019-08-20
Alias description flags


SimonBarendse:
Can we create an alias `--description` and `--descr` for this? I find myself typing it wrong every time. `--description` should also be the default if you ask me.

_Originally posted by @mackenbach in https://github.com/secrethub/secrethub-cli/pull/138_

SimonBarendse:
We should also add these aliases for the `secrethub org` command.

Currently `secrethub org init` has `--descr`, while `secrethub service init` has `--desc`. These commands should accept the same flags and have the same flag as default in the help-text.

SimonBarendse:
This has been released in [v0.26.0](https://github.com/secrethub/secrethub-cli/releases/tag/v0.26.0)

PR #141, opened 2019-08-20
Rename description flags and add aliasses

SimonBarendse:
Use `description` as the flag in the help-text and add aliases `descr` and `desc`.

### DIFF
--- a/internals/secrethub/org_init.go
+++ b/internals/secrethub/org_init.go
@@ -28,7 +28,9 @@ func NewOrgInitCommand(io ui.IO, newClient newClientFunc) *OrgInitCommand {
 func (cmd *OrgInitCommand) Register(r Registerer) {
 	clause := r.Command("init", "Initialize a new organization account.")
 	clause.Flag("name", "The name you would like to use for your organization. If not set, you will be asked for it.").SetValue(&cmd.name)
-	clause.Flag("descr", "A description (max 144 chars) for your organization so users will recognize it. If not set, you will be asked for it.").StringVar(&cmd.description)
+	clause.Flag("description", "A description (max 144 chars) for your organization so users will recognize it. If not set, you will be asked for it.").StringVar(&cmd.description)
+	clause.Flag("descr", "").Hidden().StringVar(&cmd.description)
+	clause.Flag("desc", "").Hidden().StringVar(&cmd.description)
 	registerForceFlag(clause).BoolVar(&cmd.force)
 
 	BindAction(clause, cmd.Run)
--- a/internals/secrethub/org_init.go
+++ b/internals/secrethub/org_init.go
@@ -28,7 +28,7 @@ func NewOrgInitCommand(io ui.IO, newClient newClientFunc) *OrgInitCommand {
 func (cmd *OrgInitCommand) Register(r Registerer) {
 	clause := r.Command("init", "Initialize a new organization account.")
 	clause.Flag("name", "The name you would like to use for your organization. If not set, you will be asked for it.").SetValue(&cmd.name)
-	clause.Flag("description", "A description (max 144 chars) for your organization so users will recognize it. If not set, you will be asked for it.").StringVar(&cmd.description)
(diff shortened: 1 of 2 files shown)


SimonBarendse:
Resolves #140

internals/secrethub/org_init.go
```diff
@@ -28,7 +28,9 @@ func NewOrgInitCommand(io ui.IO, newClient newClientFunc) *OrgInitCommand {
 func (cmd *OrgInitCommand) Register(r Registerer) {
 	clause := r.Command("init", "Initialize a new organization account.")
 	clause.Flag("name", "The name you would like to use for your organization. If not set, you will be asked for it.").SetValue(&cmd.name)
-	clause.Flag("descr", "A description (max 144 chars) for your organization so users will recognize it. If not set, you will be asked for it.").StringVar(&cmd.description)
+	clause.Flag("description", "A description (max 144 chars) for your organization so users will recognize it. If not set, you will be asked for it.").StringVar(&cmd.description)
```
mackenbach: Consider: so `others` will recognize it

internals/secrethub/service_init.go
```diff
@@ -117,7 +117,9 @@ func (cmd *ServiceInitCommand) Run() error {
 func (cmd *ServiceInitCommand) Register(r Registerer) {
 	clause := r.Command("init", "Create a new service account attached to a repository.")
 	clause.Arg("path", "The service account is attached to the repository in this path and when used together with --permission, an access rule is created on the directory in this path.").Required().SetValue(&cmd.path)
-	clause.Flag("desc", "A description for the service").StringVar(&cmd.description)
+	clause.Flag("description", "A description for the service").StringVar(&cmd.description)
```
mackenbach: `A description (max 144 chars) for the service so others will recognize it.`

internals/secrethub/org_init.go
```diff
@@ -28,7 +28,9 @@ func NewOrgInitCommand(io ui.IO, newClient newClientFunc) *OrgInitCommand {
...
+	clause.Flag("descr", "").Hidden().StringVar(&cmd.description)
+	clause.Flag("desc", "").Hidden().StringVar(&cmd.description)
```
mackenbach: What happens when I use `--description` AND `--desc` at the same time? Should we do something to catch that behavior or do we say that's undefined behavior?

SimonBarendse: The last provided value is used.

---

As there's no usecase to provide both flags and this behavior is not defined in the help nor the docs, I'd feel comfortable breaking this in the future (e.g. returning an error when both flags are supplied).

How do you feel about that?

mackenbach: Yes that's cool, as long as the current behavior is consistent. So if I understand you correctly `secrethub org init --description foo --desc bar` will always use `bar` as description, correct?

SimonBarendse: Yes, that's correct.

internals/secrethub/service_init.go
SimonBarendse: The description of a service is not limited to 144 chars.

SimonBarendse: It's limit is 2048 chars

SimonBarendse: How about `A description for the service to recognize it.`? It's not only for others, as it can be useful for yourself as well.

internals/secrethub/org_init.go
SimonBarendse: How about `A description (max 144 chars) for your organization to recognize it.`? It's not only for others, as it can be useful for yourself as well.

internals/secrethub/service_init.go
mackenbach: Ah ok. Euhmm, it's good to mention `others` here because it frames it as a collaborative effort. I think once about stuff for myself (or not at all), but for others I think twice about naming things so they make sense.

So I'd say: `A description for the service so others will recognize it.`

internals/secrethub/org_init.go
mackenbach: To that I say the same as below. We should aim to frame usage of SecretHub as a collaborative effort.

PR #142, opened 2019-08-21
 Add type to service ls command 

SimonBarendse:
The translation to columns is injected into the command, so that
we can use different columns in the service aws ls command, which
will be added later.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/mitchellh/mapstructure v1.1.2
 	github.com/op/go-logging v0.0.0-20160315200505-970db520ece7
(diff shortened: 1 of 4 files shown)


mackenbach:
Is this a breaking change?

SimonBarendse:
> Is this a breaking change?

@mackenbach

It's not. The output is not provided in a format that's nice for scripting (I come back to this later), but we don't break the existing format.
In this format, I can imagine that one would rely on the order of the columns in their scripts. But, as we have not changed the order of the existing columns, those scripts will still work as well.

For example `secrethub service ls | awk '{print $2}'` to get the description of the services, still works.

---

For processing in automation scripts, a JSON format might be easier. Therefore, it might be nice to optionally return the output in JSON.

With this addition, the above script would look like this: `secrethub service ls --format=json | jq .[].description`, which is also self documenting.

Let's create a ticket (and pick this up in a new PR) if we want to add this.

PR #143, opened 2019-08-21
 Add service aws ls command

SimonBarendse:
Merge #142 first


internals/secrethub/service_ls.go
```diff
@@ -67,3 +93,46 @@ func (cmd *ServiceLsCommand) Run() error {
...
+func (sw keyServiceTable) row(service *api.Service) []string {
+	return append(sw.baseServiceTable.row(service), string(service.Credential.Type))
+}
+
+type awsServiceTable struct {
+	baseServiceTable
+}
+
+func (sw awsServiceTable) header() []string {
+	return append(sw.baseServiceTable.header(), "ROLE", "KMS KEY")
```
mackenbach: Will this display the role twice? Once in the `DESCRIPTION` and once in the `ROLE` column?

internals/secrethub/service_ls.go
```diff
@@ -67,3 +93,46 @@ func (cmd *ServiceLsCommand) Run() error {
...
+}
+
+func (sw awsServiceTable) row(service *api.Service) []string {
+	return append(sw.baseServiceTable.row(service), service.Credential.Metadata[api.CredentialMetadataAWSRole], service.Credential.Metadata[api.CredentialMetadataAWSKMSKey])
```
mackenbach: `api.CredentialMetadataAWSRole` and `api.CredentialMetadataAWSKMSKey` are pretty lengthy names, starting to wonder if simply using a `aws_role` string here would be more self-documenting...

internals/secrethub/service_ls_test.go
```diff
@@ -57,6 +66,62 @@ func TestServiceLsCommand_Run(t *testing.T) {
...
+							ServiceID:   "test2",
+							Description: "foobarbaz",
+							Credential: &api.Credential{
+								Type: api.CredentialTypeRSA,
+							},
+						},
+					},
+				},
+			},
+			out: "ID    DESCRIPTION  ROLE                                   KMS KEY\ntest  foobar       arn:aws:iam::123456:role/path/to/role  arn:aws:kms:us-east-1:123456:key/12345678-1234-1234-1234-123456789012\n",
```
mackenbach: The KMS KEY ID is basically an ARN too? Does the UUID `12345678-1234-1234-1234-123456789012` suffice here?

internals/secrethub/service_ls.go
```diff
@@ -67,3 +93,46 @@ func (cmd *ServiceLsCommand) Run() error {
...
+}
+
+func isAWSService(service *api.Service) bool {
```
mackenbach: I know we're only using this internally so it won't be likely to happen, but you're accepting a pointer here and then checking a field of the struct. Seems good to have an `if service == nil { return false }` check.

internals/secrethub/service_ls.go
SimonBarendse: These keys are part of the implicit contract we have between the client and CLI. By using constants to refer to them, we make this contract a bit more explicit and less error-prone, as we ensure the same key is used in all places.

SimonBarendse: Alternatively, we could add a function in the client that translates an `api.Service` to an aws service type, which explicitly has these fields.

```Go
type AWSService struct {
	*api.Service
	Role   string
	KMSkey string
}

func ToAWSService(service *api.Service) (AWSService, error) {
	if service.Credential.Type != api.CredentialTypeAWSSTS {
		// err
	}

	role, ok := service.Credential.Metadata[api.CredentialMetadataAWSRole]
	if !ok {
		// err
	}

	kmsKey, ok := service.Credential.Metadata[api.CredentialMetadataAWSKMSKey]
	if !ok {
		// err
	}

	return AWSService{
		Service: service,
		Role:    role,
		KMSkey:  kmsKey,
	}, nil
}
```

internals/secrethub/service_ls.go
SimonBarendse: If the service is created with the CLI and no description is given, yes, as the description defaults to the role.

internals/secrethub/service_ls_test.go
SimonBarendse: I'd think the account and region are important as well.

@jpcoenen What'd be best to display here?

jpcoenen: Other places we use just the ID.

SimonBarendse: Ahh, that's actually also what is currently returned :+1: Just mocking this wrong.
Should we rename the metadata key to `kms_key_id` to make this clearer?

Do we want to add the role as another credential metadata entry then?

internals/secrethub/service_ls.go
mackenbach: Alright, let's just stick with the constants for now.

internals/secrethub/service_ls_test.go
mackenbach: > Should we rename the metadata key to kms_key_id to make clearer what's returned?

If the ID is what it is, then certainly yes make it `kms_key_id` 👍 

> Do we want to add the region as another credential metadata entry then?

That seems like a very good candidate to add yes 👍

internals/secrethub/service_ls.go
mackenbach: Okay, so think about that for a second. 

You're a user and you are creating a service and providing the role ARN. Then you run `secrethub service aws ls` and you see that long ARN printed twice. What's your initial reaction?

More confused or more like yay 🎉 ?

How can we improve the developer's user experience here?

SimonBarendse: #145

internals/secrethub/service_ls.go
```diff
@@ -67,3 +93,50 @@ func (cmd *ServiceLsCommand) Run() error {
...
+func (sw keyServiceTable) row(service *api.Service) []string {
+	return append(sw.baseServiceTable.row(service), string(service.Credential.Type))
+}
+
+type awsServiceTable struct {
+	baseServiceTable
+}
+
+func (sw awsServiceTable) header() []string {
+	return append(sw.baseServiceTable.header(), "ROLE", "KMS KEY")
```
jpcoenen: `KMS KEY` => `KMS-KEY`?

Reason is that an empty table now displays as:
```
ID  DESCRIPTION  ROLE  KMS KEY
```
Which suggests there are 5 columns instead of 4.

PR #145, opened 2019-08-27
Change the default aws service description to only the role name

SimonBarendse:


internals/secrethub/service_aws_init_test.go
```diff
@@ -0,0 +1,34 @@
+package secrethub
+
+import (
+	"github.com/secrethub/secrethub-go/internals/assert"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

internals/secrethub/service_aws_init.go
```diff
@@ -97,3 +98,22 @@ func (cmd *ServiceAWSInitCommand) Register(r Registerer) {
...
+func roleNameFromRole(role string) string {
+	if strings.Contains(role, ":") {
+		parts := strings.SplitN(role, "role/", 2)
+		if len(parts) == 2 {
+			return parts[1]
+		}
+	} else {
+		return strings.TrimPrefix(role, "role/")
+	}
+	return ""
```
mackenbach: Can we rewrite this code to be a bit easier to evaluate? E.g. moving this return to just under the `if len(parts) == 2` check on line 114 and removing the else statement on line 115? Took me a few WTFs to understand that this line is actually not dead code.

PR #147, opened 2019-08-28
Rename description flags and add aliasses

SimonBarendse:
#141, now also for the `service aws init` command.


internals/secrethub/service_aws_init.go
```diff
@@ -93,7 +93,9 @@ func (cmd *ServiceAWSInitCommand) Register(r Registerer) {
 	clause.Flag("kms-key-id", "ID of the KMS-key to be used for encrypting the service's account key.").Required().StringVar(&cmd.kmsKeyID)
 	clause.Flag("role", "ARN of the IAM role that should have access to this service account.").Required().StringVar(&cmd.role)
 	clause.Flag("region", "The AWS region that should be used").StringVar(&cmd.region)
-	clause.Flag("desc", "A description for the service. Defaults to the ARN of the IAM role that is given access to the service.").StringVar(&cmd.description)
+	clause.Flag("description", "A description for the service so others will recognize it. Defaults to the ARN of the IAM role that is given access to the service.").StringVar(&cmd.description)
```
mackenbach: No longer true is it? I think it is something like `Defaults to the IAM role name that is attached to the service.` now.

PR #154, opened 2019-09-03
Demo App

jpcoenen:
This can be used inside the Getting Started to use as a simple demo app. Can be started with `secrethub example-app serve`. It then serves a web page on the requested ports that tries to connect with httpbin.org. It only succeeds if the APP_USERNAME and APP_PASSWORD environment variables have been set to `demo` and `very-secret-password` respectively.

image

image

image


internals/example-app/serve.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	w.WriteHeader(resp.StatusCode)
+	if resp.StatusCode == http.StatusUnauthorized {
+		fmt.Fprint(w, "unauthorized")
+	} else {
+		io.Copy(w, resp.Body)
+	}
+}
+
+func writeError(w http.ResponseWriter, statusCode int, message string) {
```
golangcibot: `writeError` - `statusCode` always receives `http.StatusInternalServerError` (`500`) (from `unparam`)

internals/example-app/example.go
```diff
@@ -0,0 +1,35 @@
+package example_app
```
golangcibot: don't use an underscore in package name (from `golint`)

internals/example-app/page.go
```diff
@@ -0,0 +1,94 @@
+package example_app
```
golangcibot: don't use an underscore in package name (from `golint`)

internals/example-app/serve.go
```diff
@@ -0,0 +1,102 @@
+package example_app
```
golangcibot: don't use an underscore in package name (from `golint`)

internals/example-app/serve.go
```diff
@@ -0,0 +1,102 @@
...
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, err.Error())
+		return
+	}
+	w.WriteHeader(resp.StatusCode)
+	if resp.StatusCode == http.StatusUnauthorized {
+		fmt.Fprint(w, "unauthorized")
+	} else {
+		io.Copy(w, resp.Body)
```
golangcibot: Error return value of `io.Copy` is not checked (from `errcheck`)

mackenbach:
Question: how many secrets does the user need to write before being able to use this app? Is it both a username and a password? 

If so, could we make it even simpler by using just 1 secret, e.g. an API token? Or making the username non-secret (which I don't think we should encourage btw).

mackenbach:
Also, I love the big logo a lot. However, it is distracting me a bit from the result. Because the logo is so big and vibrant, I don't actually notice the green checkmark that much. Consider making it a bit smaller and put emphasis on the message instead?

florisvdg:
I agree with @mackenbach, let's only use the password as a secret.

internals/example-app/serve.go
```diff
@@ -0,0 +1,102 @@
...
+	host string
+	port int
+
+	appUsername string
+	appPassword string
+}
+
+// Register registers the command, arguments and flags on the provided Registerer.
+func (cmd *ServeCommand) Register(r command.Registerer) {
+	clause := r.Command("serve", "Runs the secrethub example by serving a web page.")
```
florisvdg: And, with these future guides in mind, I have a slight preference for:
`secrethub run -- secrethub example-app start` 
Over:
`secrethub run -- secrethub example-app serve` 

Matches nicer with "wrap your app start command in `secrethub run`"

internals/example-app/page.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		function processError(message){
+            processResult("An error occurred!", "times", "red", message);
+		}
+
+		$(function(){
+		    $.get("http://127.0.0.1:{{.}}/api", {}, function(res, status, xhr){
+		        if(xhr.status === 200) {
+                    processResult("Successfully connected to https://demo.secrethub.io/api!", "check", "green", res);
```
florisvdg: What's this `https://demo.secrethub.io/api`?

florisvdg:
Overall, very nice. Good idea to include this in the `secrethub-cli`. Makes deploying the example app much easier on every platform. 

It could be worthwile to make the actual example app slightly more isolated from the CLI, e.g. by putting it in a subfolder:
```
example-app/
   example.go
   app/
      serve.go # or maybe just main.go?
      page.go
      index.html
```
That way we can link to the example, and just the example, no CLI stuff. We could even add an example `Dockerfile` and `secrethub.env` in that folder for reference. Then we can also delete this repo: https://github.com/secrethub/example-app

internals/example-app/page.go
jpcoenen: It's a custom API that's in the making

jpcoenen:
> It could be worthwile to make the actual example app slightly more isolated from the CLI, e.g. by putting it in a subfolder:

@florisvdg: what do you means? It already is in a separate subfolder.

internals/example-app/serve.go
jpcoenen: The reason for choosing this command is that it:
1) Is clear about what it does: serve a page.
2) Allows you to later add different methods (like `print` or `console` which just dumps the output)

florisvdg:
Not really, because it's mixed with CLI stuff: https://github.com/secrethub/secrethub-cli/blob/feature/example-app/internals/example-app/example.go

jpcoenen:
I get what you mean now. Didn't have that sharp because I made it a month ago.

See latest commit: is this what you meant?

internals/demo-app/app/serve.go
```diff
@@ -0,0 +1,82 @@
...
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, err.Error())
+		return
+	}
+	w.WriteHeader(resp.StatusCode)
+	io.Copy(w, resp.Body)
+}
+
+func writeError(w http.ResponseWriter, statusCode int, message string) {
```
golangcibot: `writeError` - `statusCode` always receives `http.StatusInternalServerError` (`500`) (from `unparam`)

internals/demo-app/command.go
```diff
@@ -0,0 +1,25 @@
+package demo_app
```
golangcibot: don't use an underscore in package name (from `golint`)

internals/demo-app/serve.go
```diff
@@ -0,0 +1,38 @@
+package demo_app
```
golangcibot: don't use an underscore in package name (from `golint`)

internals/demo-app/app/serve.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	req.SetBasicAuth(s.appUsername, s.appPassword)
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, err.Error())
+		return
+	}
+	w.WriteHeader(resp.StatusCode)
+	io.Copy(w, resp.Body)
```
golangcibot: Error return value of `io.Copy` is not checked (from `errcheck`)

internals/secrethub/app.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	"github.com/secrethub/secrethub-cli/internals/cli"
 	"github.com/secrethub/secrethub-cli/internals/cli/ui"
+	"github.com/secrethub/secrethub-cli/internals/demo-app"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	demo_app "github.com/secrethub/secrethub-cli/internals/demo-app"
```

jpcoenen:
@florisvdg did you test it?

PR #155, opened 2019-09-04
Fix masking problem

jpcoenen:
For some inputs, the masking completely garbled the output. For example `secrethub run -- terraform plan` lead to unreadable output.

The cause of this was a wrong implementation of the `io.Writer` interface by `MaskedWriter`. It assumed control of the `[]byte` it takes as input, but it is not allowed to do so. 

The problem is demonstrated in the new `TestNewMaskedWriter/reuse_input_buffer` test.

### DIFF
--- a/internals/cli/masker/writer_test.go
+++ b/internals/cli/masker/writer_test.go
@@ -121,7 +121,7 @@ func TestMatcher(t *testing.T) {
 func TestNewMaskedWriter(t *testing.T) {
 	maskString := "<redacted by SecretHub>"
 
-	timeout20ms := time.Millisecond * 20
+	timeout10s := time.Second * 10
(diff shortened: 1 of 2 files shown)


internals/cli/masker/writer_test.go
```diff
@@ -216,6 +220,32 @@ func TestNewMaskedWriter(t *testing.T) {
 			timeout:  &timeout0,
 			expected: "test " + maskString + " test",
 		},
+		"long input": {
+			maskStrings: []string{},
+			inputFunc: func(w io.Writer) {
+
+				assert.OK(t, err)
```
SimonBarendse: What `err` is checked here?

jpcoenen: An error that rests in pieces.

PR #156, opened 2019-09-04
Add nfpms to build .deb and .rpm packages

jpcoenen:
Especially the `.rpm` is handy for AWS. Allows us to do the following for installing on AWS Linux AMI:

```
yum install -y https://github.com/secrethub/secrethub-cli/releases/latest/download/secrethub-cli-amd64.rpm
```

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -76,3 +76,16 @@ scoop:
   description: Command-line interface for SecretHub
 
   license: Apache-2.0
+
+nfpms:
+  -
+    vendor: SecretHub
+    homepage: https://secrethub.io/
+    description: CLI for using the SecretHub Secret Management Service
+    license: Apache 2.0
+    formats:
+      - rpm
+      - deb
+    recommends:
+      - xclip
+    bindir: /usr/local
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,7 +1,9 @@
 project_name: secrethub
 
 builds:
-  - binary: "secrethub"
+  - &default
+    id: default
+    binary: "bin/secrethub"
     env:
       - CGO_ENABLED=0
     goos:
@@ -33,9 +35,14 @@ builds:
       - -s -w -X "github.com/secrethub/secrethub-cli/internals/secrethub.Commit={{ .ShortCommit }}" -X "github.com/secrethub/secrethub-cli/internals/secrethub.Version={{ .Version }}"
     flags:
       - -tags=production
+  - <<: *default
+    id: "without-bin-dir"
+    binary: "secrethub"
 
 archive:
   name_template: "{{ .ProjectName }}-{{ .Tag }}-{{ .Os }}-{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}"
+  builds:
+    - default
   format_overrides:
     - goos: windows
       format: zip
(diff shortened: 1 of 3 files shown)


scripts/post-install.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env sh
+
+if [ -d /etc/bash_completion.d/ ]; then
+    echo -e "${OK_COLOR}==> Installing completion for Bash${NO_COLOR}"
```
mackenbach: Where are `${OK_COLOR}` and `${NO_COLOR}` defined?

.goreleaser.yml
```diff
@@ -1,7 +1,7 @@
 project_name: secrethub
 
 builds:
-  - binary: "bin/secrethub"
+  - binary: "secrethub"
```
mackenbach: Why did we remove this here? That seems like it could be a 'breaking' change for people that built scripts that pull the archive and place it somewhere.

florisvdg: This is breaking indeed.

mackenbach: Haha gotta love your rant. You make valid points though. What I'm fuzzy on at the moment are the reasons why we have the `bin` directory in the archive in the first place. If we can verify those reasons are no longer valid, then it can be worth it to remove the `bin` directory in favor of your proposal.

A few reasons I can think of we put it there are:
- **Because we include one or more non-binary files in the release archive**, e.g. the `LICENSE`. Specifically the `LICENSE` file we can remove and keep in the repo instead. That's fine. We need to think whether we may want to add other non-binary files in there too.
- **Because having a `bin` directory allows you to easily add the folder to your `PATH`.** Arguably, there are better ways to make the binary executable on the user's OS, like placing it in `/usr/local/bin`. However, we need to specify this as it is different per OS: macOS doesn't allow you to put files in some directories anymore I remember and Windows is always funky.
- **Because we may want to include multiple binaries in the archive.** For this I think we can make separate releases if placing them all together in the same archive proves problematic?

@SimonBarendse and @florisvdg please weigh in on this because I no longer remember all the reasons why.

SimonBarendse: I believe the main reason was that we wanted to have the license file in the archive and on the users system when installing. We did of course not want to add the license to the path, so we decided to have a bin dir instead and add that to the path.

I believe it makes sense to drop the bin dir, as we no longer require the license to be shipped and stored with every installation; It is more often unpractical than it is convenient.

If we get new non-binary files we want to ship with the release (e.g. a readme), we could add them as separate assets to the release, instead of putting them in the archive.

If we have multiple binaries to release, I think we would release these separately from each other (like we already do with the CLI, Terraform Provider and HTTP proxy). If for some reason we do want to have a shared repo with shared releases for them, we can either add different assets for these binaries, or include multiple binaries in the root of the archive.

florisvdg: Okay, but if the archive only contains the binary, we could also just directly provide the binary.

mackenbach: >Okay, but if the archive only contains the binary, we could also just directly provide the binary.

Archive formats are nice because of three things:

- Compression (duh)
- Avoids nasty warnings about downloading a binary/.exe
- People expect it.

jpcoenen: And I can add one more benefit: my archive manager has this awesome option.

image

PR #157, opened 2019-09-04
Use the region from the KMS key ARN when creating AWS services

SimonBarendse:
This is used when no region is set in the aws configuration or with
the --region flag.

If the KMS key is not given as an ARN, the user will still be prompted
for the region.

### DIFF
--- a/internals/secrethub/service_aws_init.go
+++ b/internals/secrethub/service_aws_init.go
@@ -86,19 +86,21 @@ func (cmd *ServiceAWSInitCommand) Run() error {
 
 	fmt.Fprintf(cmd.io.Stdout(), "Detected access to AWS account %s.", accountID)
 
-	if !isSet(cfg.Region) && cmd.kmsKeyID != "" {
+	if cfg.Region == nil && cmd.kmsKeyID != "" {
+		// When the region is not configured in the AWS configuration and not supplied using the flag, use
+		// the region from the KMS key if the key is supplied as an ARN.
 		kmsARN, err := arn.Parse(cmd.kmsKeyID)
 		if err == nil {
 			cfg = cfg.WithRegion(kmsARN.Region)
 		}
 	}
 
-	if isSet(cfg.Region) {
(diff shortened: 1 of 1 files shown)


internals/secrethub/service_aws_init.go
```diff
@@ -85,13 +86,19 @@ func (cmd *ServiceAWSInitCommand) Run() error {
...
+		kmsARN, err := arn.Parse(cmd.kmsKeyID)
+		if err == nil {
+			cfg = cfg.WithRegion(kmsARN.Region)
+		}
+	}
+
+	if isSet(cfg.Region) {
+		fmt.Fprintf(cmd.io.Stdout(), "Using region %s.", *cfg.Region)
 	}
 	fmt.Fprintln(cmd.io.Stdout())
```
mackenbach: This logic is a bit hard to understand. If it is set, you print something, then you print a newline always and else you ask for input (printing again). Is there a way to make this easier to read?

internals/secrethub/service_aws_init.go
```diff
@@ -85,13 +86,19 @@ func (cmd *ServiceAWSInitCommand) Run() error {
 
 	fmt.Fprintf(cmd.io.Stdout(), "Detected access to AWS account %s.", accountID)
 
-	region := sess.Config.Region
-	if isSet(region) {
-		fmt.Fprintf(cmd.io.Stdout(), "Using region %s.", *region)
+	if !isSet(cfg.Region) && cmd.kmsKeyID != "" {
```
mackenbach: Consider adding a comment here that explains what it does. It's a bit confusing to read. There are multiple `if isSet` and `if !isSet` statements here and there is an additional condition `if err != nil` inside here that is intentionally ignored.

florisvdg:
And about the PR title... 😁https://www.youtube.com/watch?v=PEgk2v6KntY

SimonBarendse:
@florisvdg Way better :ok_hand:

PR #159, opened 2019-09-06
Add optional argument to repo ls to limit results to repos within a workspace

SimonBarendse:
We can later alias the `ls <workspace>` command to this command.


SimonBarendse:
@mackenbach Yes, we can add that as well. I intend to do so by aliasing to this command.

SimonBarendse:
@mackenbach I've opened #162 to add `ls <workspace>` support.

PR #164, opened 2019-09-09
 Ask to create an organization on signup

SimonBarendse:


internals/secrethub/signup.go
```diff
@@ -160,6 +176,15 @@ func (cmd *SignUpCommand) Run() error {
 		return err
 	}
 
+	if cmd.org != "" {
+		_, err := client.Orgs().Create(cmd.org, cmd.orgDescription)
+		if err == api.ErrOrgAlreadyExists {
+			fmt.Fprintln(cmd.io.Stdout(), "The organization already exists. Ask a colleague to invite you to the organization.")
```
SimonBarendse: We can improve this behavior in future releases (e.g. request access)

internals/secrethub/signup.go
```diff
@@ -109,6 +113,18 @@ func (cmd *SignUpCommand) Run() error {
 					return err
 				}
 			}
+			if cmd.org == "" {
+				cmd.org, err = ui.AskAndValidate(cmd.io, "Your organization (leave empty to skip for now): ", 2, allowEmpty(api.ValidateOrgName))
```
florisvdg: What about:
```
Would you like to create a workspace for your organization right away? 
(This can be done later by running secrethub org init) [Y/n]
The name of your organization workspace:
```

SimonBarendse: Why?

We'd like to encourage users to create the organization right away. With the proposed change, users have to answer an extra question to get there.
Users that do not want to create an organization, now also require more input (`n` + `ENTER`).

florisvdg: To prevent a support flood for reclaims of dangling orgs.

SimonBarendse: Oke, in that case, I think we should rethink whether we want to encourage creating the org right away in the first place.

Note that the workspace name (not the secrets) can be reclaimed even when access to the credential is lost. The flow is not yet optimal, but one could request a password reset via the dashboard and connect a CLI via `account init`, the account then has access to all namespaces it had before. Access to directories and secrets cannot be reset without a credential, as they are encrypted with the credential.

mackenbach: First off, we discussed this at length before: yes we definitely want to encourage creating an organization on signup. It sets the stage for collaboration, very important for customer success.  

Second, let's take a page from the Slack playbook here for the copy: https://www.useronboard.com/how-slack-onboards-new-users/?slide=36

Something like this could work: 

- `Choose a shared workspace name for your team (e.g. your company, leave empty to skip for now):`
- `Team workspace name, e.g. your company (leave empty to skip):`
- `Team workspace name, usually your company name (leave empty to skip):`

SimonBarendse: I like this one: `Team workspace name, usually your company name (leave empty to skip):`

Following the advice supplied in the guide you provided, should we provide extra context? e.g.

`Team workspace name (this is where your shared repositories will live), usually your company name (leave empty to skip):`

florisvdg: `Workspace name for your team (e.g. your company, leave empty to skip for now):`

mackenbach: Hypothesis: I think the concept of `workspace` is fairly familiar in the minds of our new users.

So IMO it's fine to leave out the `this is where your shared repositories will live` part. One of the feedback points we god for the signup flow was that it was a lot of text (especially the lines about where the credential file is stored, which we moved to the docs I believe), so I'd advocate for brief descriptive questions whenever possible. It's already a long-ish one.

SimonBarendse: Now that we've changed the prompt to say workspace and nothing about an organization, should we change the flag names and descriptions too, to refer to team workspaces rather than organizations?

internals/secrethub/signup.go
```diff
@@ -181,7 +185,37 @@ func (cmd *SignUpCommand) Run() error {
...
+	if !createWorkspace {
+		createWorkspace, err = ui.AskYesNo(cmd.io, "Do you want to create a shared workspace for your team?", ui.DefaultYes)
+		if err != nil {
+			return err
+		}
+	}
+	if createWorkspace {
+		fmt.Fprint(cmd.io.Stdout(), "Please answer the questions below, followed by an [ENTER]\n\n")
+		if cmd.org == "" {
+			cmd.orgDescription, err = ui.AskAndValidate(cmd.io, "workspace name for your team (e.g. your company name): ", 2, api.ValidateOrgName)
```
mackenbach: You're setting `cmd.orgDescription` here while it's the `cmd.org` field that needs to be set.

internals/secrethub/signup.go
```diff
@@ -181,7 +185,37 @@ func (cmd *SignUpCommand) Run() error {
 	}
 
 	cmd.progressPrinter.Stop()
-
 	fmt.Fprintf(cmd.io.Stdout(), "Setup complete. To read your first secret, run:\n\n    secrethub read %s\n\n", secretPath)
```
mackenbach: I think this line should be printed after the entire thing is complete. now it gets lost in translation. so the flow becomes:

1. questions for account
2. questions for org
3. create account
4. create org
5. success message / helpful error message if org already exists

internals/secrethub/signup.go
```diff
@@ -181,7 +185,37 @@ func (cmd *SignUpCommand) Run() error {
...
+
+	createWorkspace := cmd.org != ""
+	if !createWorkspace {
+		createWorkspace, err = ui.AskYesNo(cmd.io, "Do you want to create a shared workspace for your team?", ui.DefaultYes)
+		if err != nil {
+			return err
+		}
+	}
+	if createWorkspace {
+		fmt.Fprint(cmd.io.Stdout(), "Please answer the questions below, followed by an [ENTER]\n\n")
```
mackenbach: As said in the comment above, I think the flow needs to be changed. If you do that, then you don't need this additional line. It becomes self explanatory.

internals/secrethub/signup.go
SimonBarendse: oops

internals/secrethub/signup.go
SimonBarendse: We discussed last Friday that separating the creation of the account and the org workspace gives us two advantages:
1. It keeps the signup part simple and straightforward and adds the shared workspace as an optional extra goodie. We don't want to create extra hurdles in signing up.
1. It makes messages easier and clearer, as they're given in the context of the phase they're printed in. "workspace already exists" in the account creation phase means the username already exists, in the shared workspace creation phase it means the orgname already exists.


To solve the problem of the "to read your first secret ..." being printed before the command has finished. How about printing "Setup complete." where it is right now, but move the "to read your first secret ..." to the end of the command?

mackenbach: I feel we mean the same thing but it looks like we're disagreeing. Lets discuss this in person, e.g. tomorrow.

internals/secrethub/signup.go
```diff
@@ -43,9 +45,11 @@ func NewSignUpCommand(io ui.IO, newClient newClientFunc, credentialStore Credent
 // Register registers the command, arguments and flags on the provided Registerer.
 func (cmd *SignUpCommand) Register(r command.Registerer) {
 	clause := r.Command("signup", "Create a free personal developer account.")
-	clause.Flag("username", "The username you would like to use on SecretHub. If not set, you will be asked for it.").StringVar(&cmd.username)
-	clause.Flag("full-name", "If not set, you will be asked to provide your full name.").StringVar(&cmd.fullName)
-	clause.Flag("email", "If not set, you will be asked to provide your email address.").StringVar(&cmd.email)
+	clause.Flag("username", "The username you would like to use on SecretHub.").StringVar(&cmd.username)
+	clause.Flag("full-name", "Your email address.").StringVar(&cmd.fullName)
```
mackenbach: Make it `Your (work) email address`, also in the prompt

internals/secrethub/signup.go
```diff
@@ -181,7 +185,42 @@ func (cmd *SignUpCommand) Run() error {
...
+		if err != nil {
+			return err
+		}
+		if !createWorkspace {
+			fmt.Fprintln(cmd.io.Stdout(), "You can create a shared workspace later using `secrethub org init`.")
+		}
+	}
+	if createWorkspace {
+		if cmd.org == "" {
+			cmd.org, err = ui.AskAndValidate(cmd.io, "Workspace name for your team (e.g. your company name): ", 2, api.ValidateOrgName)
```
mackenbach: the `for your team` part is a bit redundant as it's already stated in the question above `do you want to create a shared workspace for your team?`. Removing it here makes it a bit shorter:

`Workspace name (e.g. your company name):`

internals/secrethub/signup.go
```diff
@@ -181,7 +185,42 @@ func (cmd *SignUpCommand) Run() error {
...
+		_, err := client.Orgs().Create(cmd.org, cmd.orgDescription)
+		if err == api.ErrOrgAlreadyExists {
+			fmt.Fprintf(cmd.io.Stdout(), "The workspace %s already exists. If it is your organization, ask a colleague to invite you to the workspace. You can also create a new one using `secrethub org init`.\n", cmd.org)
+		} else if err != nil {
+			return err
+		} else {
+			fmt.Fprintln(cmd.io.Stdout(), "Created your shared workspace.")
+		}
+	}
+	fmt.Fprintf(cmd.io.Stdout(), "\nTo read your first secret, run:\n\n    secrethub read %s\n\n", secretPath)
```
mackenbach: Do we want to say here: `Setup complete. To read your first secrets ...`?

SimonBarendse: We either have:
```
Created your account.
Do you want to create a shared workspace for your team? (Y/n) no

To read your first secret, run:

secrethub read your-username/start/hello
```
or:
```
Created your account.
Do you want to create a shared workspace for your team? (Y/n) yes
Workspace name for your team (e.g. your company name): company-name
A description (max 144 chars) for your team workspace so others will recognize it:
This is my company.
Created your shared workspace.

To read your first secret, run:

secrethub read your-username/start/hello
```

Yes, maybe that makes sense to add, to indicate the shared workspace was the last step :+1:

mackenbach: Yes and add a white line between the question block, so it becomes something like:
```
Created your account.
Do you want to create a shared workspace for your team? (Y/n) yes

Workspace name for your team (e.g. your company name): company-name
A description (max 144 chars) for your team workspace so others will recognize it:
This is my company.

Created your shared workspace.
To read your first secret, run:

secrethub read your-username/start/hello
```

mackenbach:
@SimonBarendse what's blocking this from being fixed and merged?

PR #171, opened 2019-09-23
 Rename generate rand to generate and make length a flag 

SimonBarendse:
New: `generate [--length=<length>] <path>`
Old: `generate rand <path> [<length>]` still works.

### DIFF
--- a/internals/secrethub/generate.go
+++ b/internals/secrethub/generate.go
@@ -2,9 +2,12 @@ package secrethub
 
 import (
 	"fmt"
+	"strconv"
+
+	"github.com/secrethub/secrethub-go/internals/api"
 
 	"github.com/secrethub/secrethub-cli/internals/cli/ui"
-	"github.com/secrethub/secrethub-go/internals/api"
+
 	"github.com/secrethub/secrethub-go/internals/errio"
 	"github.com/secrethub/secrethub-go/pkg/randchar"
 )
@@ -13,16 +16,21 @@ var (
 	errGenerate = errio.Namespace("generate")
 
 	// ErrInvalidRandLength is returned when an invalid length is given.
-	ErrInvalidRandLength = errGenerate.Code("invalid_rand_length").Error("The secret length must be larger than 0")
+	ErrInvalidRandLength         = errGenerate.Code("invalid_rand_length").Error("The secret length must be larger than 0")
+	ErrCannotUseLengthArgAndFlag = errGenerate.Code("length_arg_and_flag").Error("length cannot be provided as an argument and a flag at the same time")
 )
+
+const defaultLength = 22
 
 // GenerateSecretCommand generates a new secret and writes to the output path.
 type GenerateSecretCommand struct {
 	useSymbols bool
 	generator  randchar.Generator
 	io         ui.IO
-	length     int
-	path       api.SecretPath
+	lengthFlag *int
+	firstArg   string
+	secondArg  string
+	lengthArg  *int
 	newClient  newClientFunc
 }
 
@@ -36,12 +44,14 @@ func NewGenerateSecretCommand(io ui.IO, newClient newClientFunc) *GenerateSecret
 
 // Register registers the command, arguments and flags on the provided Registerer.
 func (cmd *GenerateSecretCommand) Register(r Registerer) {
-	generateCommand := r.Command("generate", "").Hidden()
-
-	clause := generateCommand.Command("rand", "Generate a random secret. By default, it uses numbers (0-9), lowercase letters (a-z) and uppercase letters (A-Z) and a length of 22.")
-	clause.Arg("secret-path", "The path to write the generated secret to (<namespace>/<repo>[/<dir>]/<secret>)").Required().SetValue(&cmd.path)
-	clause.Arg("length", "The length of the generated secret. Defaults to 22.").Default("22").IntVar(&cmd.length)
+	clause := r.Command("generate", "Generate a random secret.")
+	clause.HelpLong("By default, it uses numbers (0-9), lowercase letters (a-z) and uppercase letters (A-Z) and a length of 22.")
+	clause.Arg("secret-path", "The path to write the generated secret to (<namespace>/<repo>[/<dir>]/<secret>)").Required().StringVar(&cmd.firstArg)
+	cmd.lengthFlag = clause.Flag("length", "The length of the generated secret. Defaults to "+strconv.Itoa(defaultLength)).PlaceHolder(strconv.Itoa(defaultLength)).Int()
 	clause.Flag("symbols", "Include symbols in secret.").Short('s').BoolVar(&cmd.useSymbols)
+
+	clause.Arg("rand-command", "").Hidden().StringVar(&cmd.secondArg)
+	cmd.lengthArg = clause.Arg("length", "").Hidden().Int()
 
 	// TODO SHDEV-528: implement --clip
 	// clause.Flag("clip", "Copy the secret value to the clipboard. The clipboard is automatically cleared after 45 seconds.").Short('c').BoolVar(cmd.clip)
@@ -62,13 +72,23 @@ func (cmd *GenerateSecretCommand) Run() error {
 
 // run generates a new secret and writes to the output path.
 func (cmd *GenerateSecretCommand) run() error {
-	if cmd.length <= 0 {
+	length, err := cmd.length()
+	if err != nil {
+		return err
+	}
+
+	path, err := cmd.path()
+	if err != nil {
+		return err
+	}
+
+	if length <= 0 {
 		return ErrInvalidRandLength
 	}
 
 	fmt.Fprint(cmd.io.Stdout(), "Generating secret value...\n")
 
-	data, err := cmd.generator.Generate(cmd.length)
+	data, err := cmd.generator.Generate(length)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


internals/secrethub/generate_test.go
```diff
@@ -49,27 +53,109 @@ func TestGenerateSecretCommand_run(t *testing.T) {
...
+				secondArg: "namespace/repo/secret",
+				lengthArg: newIntValue(-1),
 			},
 			err: ErrInvalidRandLength,
 			out: "",
 		},
+		// The length arg is only for backwards compatibility of the `generate rand` command.
+		"length arg without rand": {
+			cmd: GenerateSecretCommand{
+				firstArg: "namespace/repo/secret",
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

internals/secrethub/generate.go
```diff
@@ -13,16 +15,21 @@ var (
 	errGenerate = errio.Namespace("generate")
 
 	// ErrInvalidRandLength is returned when an invalid length is given.
-	ErrInvalidRandLength = errGenerate.Code("invalid_rand_length").Error("The secret length must be larger than 0")
+	ErrInvalidRandLength         = errGenerate.Code("invalid_rand_length").Error("The secret length must be larger than 0")
+	ErrCannotUseLengthArgAndFlag = errGenerate.Code("length_arg_and_flag").Error("length cannot be provided as an argument and a flag at the same time")
```
mackenbach: Would we want to be more accepting and have the `--length` flag take precedence instead of erroring?

internals/secrethub/generate.go
```diff
@@ -36,13 +43,15 @@ func NewGenerateSecretCommand(io ui.IO, newClient newClientFunc) *GenerateSecret
...
+	clause.HelpLong("By default, it uses numbers (0-9), lowercase letters (a-z) and uppercase letters (A-Z) and a length of 22.")
+	clause.Arg("secret-path", "The path to write the generated secret to (<namespace>/<repo>[/<dir>]/<secret>)").Required().StringVar(&cmd.firstArg)
+	clause.Flag("length", "The length of the generated secret. Defaults to "+strconv.Itoa(defaultLength)).PlaceHolder(strconv.Itoa(defaultLength)).SetValue(&cmd.lengthFlag)
 	clause.Flag("symbols", "Include symbols in secret.").Short('s').BoolVar(&cmd.useSymbols)
 
+	clause.Arg("rand-command", "").Hidden().StringVar(&cmd.secondArg)
+	clause.Arg("length", "").Hidden().SetValue(&cmd.lengthArg)
+
 	// TODO SHDEV-528: implement --clip
 	// clause.Flag("clip", "Copy the secret value to the clipboard. The clipboard is automatically cleared after 45 seconds.").Short('c').BoolVar(cmd.clip)
```
mackenbach: Spotted this, created an issue for it #172

internals/secrethub/generate.go
```diff
@@ -36,13 +43,15 @@ func NewGenerateSecretCommand(io ui.IO, newClient newClientFunc) *GenerateSecret
...
 func (cmd *GenerateSecretCommand) Register(r Registerer) {
-	generateCommand := r.Command("generate", "").Hidden()
-
-	clause := generateCommand.Command("rand", "Generate a random secret. By default, it uses numbers (0-9), lowercase letters (a-z) and uppercase letters (A-Z) and a length of 22.")
-	clause.Arg("secret-path", "The path to write the generated secret to (<namespace>/<repo>[/<dir>]/<secret>)").Required().SetValue(&cmd.path)
-	clause.Arg("length", "The length of the generated secret. Defaults to 22.").Default("22").IntVar(&cmd.length)
+	clause := r.Command("generate", "Generate a random secret.")
+	clause.HelpLong("By default, it uses numbers (0-9), lowercase letters (a-z) and uppercase letters (A-Z) and a length of 22.")
+	clause.Arg("secret-path", "The path to write the generated secret to (<namespace>/<repo>[/<dir>]/<secret>)").Required().StringVar(&cmd.firstArg)
+	clause.Flag("length", "The length of the generated secret. Defaults to "+strconv.Itoa(defaultLength)).PlaceHolder(strconv.Itoa(defaultLength)).SetValue(&cmd.lengthFlag)
```
mackenbach: Can we add the short `-l` as well?

internals/secrethub/generate.go
```diff
@@ -62,13 +71,23 @@ func (cmd *GenerateSecretCommand) Run() error {
...
+		return err
+	}
+
+	path, err := cmd.path()
+	if err != nil {
+		return err
+	}
+
+	if length <= 0 {
 		return ErrInvalidRandLength
```
mackenbach: Should it error in this case or just take the default?

internals/secrethub/generate.go
```diff
@@ -80,12 +99,66 @@ func (cmd *GenerateSecretCommand) run() error {
 
 	fmt.Fprint(cmd.io.Stdout(), "Writing secret value...\n")
```
mackenbach: I'm seeing this now, your call to pick it up or not, but why are we printing out `generating...writing...write complete`? If we want to print anything out, would `generating... generate complete` be sufficient?

internals/secrethub/generate.go
```diff
@@ -36,13 +43,15 @@ func NewGenerateSecretCommand(io ui.IO, newClient newClientFunc) *GenerateSecret
...
-	clause := generateCommand.Command("rand", "Generate a random secret. By default, it uses numbers (0-9), lowercase letters (a-z) and uppercase letters (A-Z) and a length of 22.")
-	clause.Arg("secret-path", "The path to write the generated secret to (<namespace>/<repo>[/<dir>]/<secret>)").Required().SetValue(&cmd.path)
-	clause.Arg("length", "The length of the generated secret. Defaults to 22.").Default("22").IntVar(&cmd.length)
+	clause := r.Command("generate", "Generate a random secret.")
+	clause.HelpLong("By default, it uses numbers (0-9), lowercase letters (a-z) and uppercase letters (A-Z) and a length of 22.")
+	clause.Arg("secret-path", "The path to write the generated secret to (<namespace>/<repo>[/<dir>]/<secret>)").Required().StringVar(&cmd.firstArg)
+	clause.Flag("length", "The length of the generated secret. Defaults to "+strconv.Itoa(defaultLength)).PlaceHolder(strconv.Itoa(defaultLength)).SetValue(&cmd.lengthFlag)
 	clause.Flag("symbols", "Include symbols in secret.").Short('s').BoolVar(&cmd.useSymbols)
 
+	clause.Arg("rand-command", "").Hidden().StringVar(&cmd.secondArg)
```
mackenbach: One thing I'm missing is: how are environment variables handled in a backwards compatible way? And, did you 'add' more environment variables that `secrethub printenv -v` will print now?

I'd suggest to only keep `SECRETHUB_GENERATE_RAND_XXX` for backwards compatibility and always have `SECRETHUB_GENERATE_XXX` take precedence.

internals/secrethub/generate.go
SimonBarendse: I'm not sure whether that's better or whether that leads to unexpected behavior?

Consider e.g. this usage:
```
SECRETHUB_GENERATE_LENGTH=12
...
secrethub generate rand path/to/secret 30
```

When the environment variable is set at another place than the argument, it might not be clear that the argument is overridden.
In all other cases, values set as a flag/arg take precedence over environment variables. So one might expect that just looking at the command (`secrethub generate rand path/to/secret 30`) is enough.

mackenbach: Ah yes, that is confusing and unpredictable indeed. Let's stick with erroring for now then. We can always be more forgiving later and we're going to drop this stuff eventually anyway.

internals/secrethub/generate.go
SimonBarendse: Good catch! For hidden flags an environment variable is still registered, but not for renamed commands. I've added a commit to read the envvar in the command.

internals/secrethub/generate.go
SimonBarendse: As this is unrelated to this PR, I've created #173 to pick this up separately.

internals/secrethub/generate.go
```diff
@@ -59,13 +60,31 @@ func (cmd *GenerateSecretCommand) Register(r Registerer) {
...
+func (cmd *GenerateSecretCommand) before() error {
+	useSymbols := false
+
+	useSymbolsEnv := os.Getenv("SECRETHUB_GENERATE_RAND_SYMBOLS")
+	if useSymbolsEnv != "" {
+		b, err := strconv.ParseBool(useSymbolsEnv)
+		if err != nil {
+			return err
+		}
+		useSymbols = useSymbols || b
```
mackenbach: What's happening here? It's already set to false, so can we do without the `||` check here and just say `useSymbols = b`?

internals/secrethub/generate.go
```diff
@@ -59,13 +60,31 @@ func (cmd *GenerateSecretCommand) Register(r Registerer) {
...
+	}
+
+	useSymbols = useSymbols || cmd.useSymbols
```
mackenbach: How does this work when the `--symbols` flag is set to `false`, but the `SECRETHUB_GENERATE_RAND_SYMBOLS` envar is set to `true`?

Reason I ask is twofold:
- First, the multiple conditionals make it a bit hard to reason about it so I had to read it a few times to understand what's happening. 
- Second, it seems to me like it may not follow the order of precedence. IMO the order of precedence should be:
   1. `--symbols` flag 
   2. `SECRETHUB_GENERATE_SYMBOLS` environment variable
   3. `SECRETHUB_GENERATE_RAND_SYMBOLS` environment variable

PR #175, opened 2019-09-24
Rename generate --symbols flag to --use-symbols

SimonBarendse:
--symbols is still accepted for backwards compatibility.


SimonBarendse:
#177

PR #180, opened 2019-09-26
Release v0.29.0

SimonBarendse:
### Added
- Add deb and rpm packages to releases
- Add `generate --clip` flag to copy the generated value 
- Rename `generate rand` to `generate` and add `generate --length` flag
- Make generate command less verbose
- Rephrase `service init --permission` helptext 

### Deprecated
- Deprecated `generate rand <path> [<length>]`, use `generate [--length <length>] <path>` instead.


.goreleaser.yml
```diff
@@ -76,3 +87,23 @@ scoop:
   description: Command-line interface for SecretHub
 
   license: Apache-2.0
+
+nfpms:
+  -
+    builds:
+      - without-bin-dir
+    vendor: SecretHub
+    homepage: https://secrethub.io/
```
florisvdg: `secrethub.io` or `secrethub.io/`? Let's make it consistent with the other urls in this file.

SimonBarendse: It's `https://secrethub.io`. Changed it.

PR #186, opened 2019-09-30
Fix bug MaskedWriter is not flushed before timeout

jpcoenen:
There was a bug in MaskedWriter where it was only flushed on a timeout when a secret was found in the middle of a write. The reason was that `matchInProgress := false` had to be inside the loop instead of outside. Also added a test to test whether the problem is solved.

### DIFF
--- a/internals/cli/masker/writer_test.go
+++ b/internals/cli/masker/writer_test.go
@@ -11,6 +11,8 @@ import (
 	"github.com/secrethub/secrethub-go/pkg/randchar"
 )
 
+var maskString = "<redacted by SecretHub>"
+
 func TestMatcher(t *testing.T) {
 	tests := []struct {
 		matchString     string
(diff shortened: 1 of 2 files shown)


internals/cli/masker/writer_test.go
```diff
@@ -271,6 +271,35 @@ func TestNewMaskedWriter(t *testing.T) {
 	}
 }
 
+func TestNewMaskedWriter_OnlyFlushingOnTimeoutBug(t *testing.T) {
```
SimonBarendse: Every test asserts that there is not a bug. Therefore, I think we should explain the test instead of the bug in the function name.

SimonBarendse: e.g. `FlushBeforeTimeout`

jpcoenen: :heavy_check_mark:

PR #187, opened 2019-10-01
Explicitly set brew install field

jpcoenen:
Because the automatically generated value is no good when using multiple
builds.


mackenbach:
@jpcoenen and @florisvdg we may want to create a CLI release and will need to include this fix. Any reason this hasn't been merged yet?

PR #190, opened 2019-10-08
Add --in-file option to write command

### LINKED ISSUE #35: Add --in-file flag to write command
Not every environment handles piping nicely and we cannot predict how users will use our CLI. For instance, a docker container with the `secrethub` binary as entrypoint cannot accept piped content (without workarounds).

## Proposed docs changes

The `write` docs would change by adding this flag:

```
-i, --input-file (string)

   Use the contents of a file as secret input. 
```

And the order in which the command reads the secret from different inputs would change to:

1. `--input-file` OR `--clip`. These two are mutually exclusive and should error when both are set.
1. If input is piped, read from `stdin`
1. If no other input is given, ask the user to type in the secret

SimonBarendse:
Write the contents of a file to a secret with the --in-file flag.

Resolves #35

### DIFF
--- a/internals/secrethub/write.go
+++ b/internals/secrethub/write.go
@@ -14,12 +14,14 @@ import (
 var (
 	errCannotWriteToVersion = errMain.Code("cannot_write_version").Error("cannot (over)write a specific secret version, they are append only")
 	errEmptySecret          = errMain.Code("cannot_write_empty_secret").Error("secret is empty or contains only whitespace")
+	errClipAndInFile        = errMain.Code("clip_and_in-file").Error("clip and in-file cannot be used together")
 )
 
 // WriteCommand is a command to write content to a secret.
 type WriteCommand struct {
 	io           ui.IO
 	path         api.SecretPath
+	inFile       string
 	useClipboard bool
 	noTrim       bool
 	clipper      clip.Clipper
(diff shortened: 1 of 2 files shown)


internals/secrethub/write_test.go
```diff
@@ -191,6 +191,13 @@ func TestWriteCommand_Run(t *testing.T) {
 			},
 			err: clip.ErrCannotRead("read error"),
 		},
+		"clip and in-file": {
+			cmd: WriteCommand{
+				inFile: "file",
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
				inFile:       "file",
```

internals/secrethub/write.go
```diff
@@ -14,12 +14,14 @@ import (
 var (
 	errCannotWriteToVersion = errMain.Code("cannot_write_version").Error("cannot (over)write a specific secret version, they are append only")
 	errEmptySecret          = errMain.Code("cannot_write_empty_secret").Error("secret is empty or contains only whitespace")
+	errClipAndInFile        = errMain.Code("clip_and_in-file").Error("clip and in-file cannot be used together")
```
florisvdg: Multicased error code is a bit ugly imo.

PR #191, opened 2019-10-08
Add --out-file and --file-mode flags to read

SimonBarendse:
Use --out-file (and --file-mode) to write the contents of a secret directly to a file.

### DIFF
--- a/internals/secrethub/read.go
+++ b/internals/secrethub/read.go
@@ -2,15 +2,22 @@ package secrethub
 
 import (
 	"fmt"
+	"io/ioutil"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internals/secrethub/read.go
```diff
@@ -60,6 +71,10 @@ func (cmd *ReadCommand) Run() error {
 		return err
 	}
 
+	if cmd.useClipboard && cmd.outFile != "" {
```
florisvdg: Why is this mutually exclusive?

SimonBarendse: Hmm, you're right. I've updated it too allow both to be used together.

PR #195, opened 2019-10-10
Use relative url in demo app

florisvdg:
So it works locally and remotely.


jpcoenen:
You can also remove the go template stuff now.

PR #200, opened 2019-10-11
YUM release changes

jpcoenen:
Some minor changes to `.goreleaser` to allow release to yum repository.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -107,4 +107,12 @@ nfpms:
     scripts:
       postinstall: "scripts/post-install.sh"
       postremove: "scripts/post-remove.sh"
+    overrides:
+      rpm:
+        replacements:
+          amd64: x86_64
+          386: i386
+        name_template: "{{ .ProjectName }}-{{ .Version }}-1.{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}"
+      deb:
+        name_template: "{{ .ProjectName }}-{{ .Tag }}-{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}"
 
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,4 +1,4 @@
-project_name: secrethub
+project_name: secrethub-cli
 
 builds:
   - &default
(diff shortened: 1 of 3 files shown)


mackenbach:
@jpcoenen @florisvdg what's the priority of this? Should we merge this before doing a next release of the CLI? If so, let's prioritize.

florisvdg:
We can also wait with it until we split up the goreleaser file for CI. That way we're sure it can't break anything.

mackenbach:
> It means the conventional names for rpm's are used. But... it is not strictly necessary. Looking at it again I'm not sure whether it is nice to rename it just for this single packaging format.

I have very little clue as to how much value this has. We're hosting our own YUM repo correct? Then renaming those files into the ones fit for the yum repo when you push to the repo (which is on s3?) sounds fine to me. 

However, making the `.rpm` files match the `.rpm` convention is also pretty good if you ask me. 

Long story short:

jpcoenen:
> I have very little clue as to how much value this has. We're hosting our own YUM repo correct? Then renaming those files into the ones fit for the yum repo when you push to the repo (which is on s3?) sounds fine to me.
> 
> However, making the `.rpm` files match the `.rpm` convention is also pretty good if you ask me.
> 
> Long story short:

YUM does not care. It will only call our `amd64` binary `X86_64` when you try to install it, regardless of the filename (`secrethub-cli X86_64 v1.2.3``)

mackenbach:
If yum don't care, then let's stay consistent with our own stuff right?

jpcoenen:
Changed it to be consistent with the archives:
```
dist/secrethub-v0.30.0-darwin-amd64.tar.gz
dist/secrethub-v0.30.0-linux-386.deb
dist/secrethub-v0.30.0-linux-386.rpm
dist/secrethub-v0.30.0-linux-386.tar.gz
dist/secrethub-v0.30.0-linux-amd64.deb
dist/secrethub-v0.30.0-linux-amd64.rpm
dist/secrethub-v0.30.0-linux-amd64.tar.gz
dist/secrethub-v0.30.0-linux-arm64.deb
dist/secrethub-v0.30.0-linux-arm64.rpm
dist/secrethub-v0.30.0-linux-arm64.tar.gz
dist/secrethub-v0.30.0-linux-armv5.deb
dist/secrethub-v0.30.0-linux-armv5.rpm
dist/secrethub-v0.30.0-linux-armv5.tar.gz
dist/secrethub-v0.30.0-linux-armv6.deb
dist/secrethub-v0.30.0-linux-armv6.rpm
dist/secrethub-v0.30.0-linux-armv6.tar.gz
dist/secrethub-v0.30.0-linux-armv7.deb
dist/secrethub-v0.30.0-linux-armv7.rpm
dist/secrethub-v0.30.0-linux-armv7.tar.gz
dist/secrethub-v0.30.0-windows-386.zip
dist/secrethub-v0.30.0-windows-amd64.zip
```

mackenbach:
Hmm.. I like how it is now, but we've already released a version of the CLI with a different structure. I'm okay(ish) with 'breaking' this once, but we should be careful and update all the docs and links that refer to it too.

jpcoenen:
As far as I know, all links to it are dynamic. So that should not be a problem.

PR #201, opened 2019-10-14
Use compound path notation in help-text

SimonBarendse:


internals/secrethub/path_placeholders.go
```diff
@@ -0,0 +1,10 @@
+package secrethub
+
+const (
+	repoPathPlaceHolder                  = "<namespace>/<repo>"
+	dirPathPlaceHolder                   = repoPathPlaceHolder + "/<dir>[/<dir> ...]"
+	optionalDirPathPlaceHolder           = repoPathPlaceHolder + "[/<dir> ...]"
+	secretPathPlaceHolder                = optionalDirPathPlaceHolder + "/<secret>"
+	secretPathOptionalVersionPlaceHolder = secretPathPlaceHolder + "[:<version>]"
+	anyPathPlaceHolder                   = "<namespace>[/<repo>[/<dir> ...][/<secret>[:<version>]]]"
```
jpcoenen: Though this is technically correct, I don't think it makes things a lot more understandable. Can we make this simpler?

internals/secrethub/repo_init.go
```diff
@@ -27,7 +27,7 @@ func NewRepoInitCommand(io ui.IO, newClient newClientFunc) *RepoInitCommand {
 // Register registers the command, arguments and flags on the provided Registerer.
 func (cmd *RepoInitCommand) Register(r command.Registerer) {
 	clause := r.Command("init", "Initialize a new repository.")
-	clause.Arg("repo-path", "Path to the new repository (<namespace>/<repo>)").Required().SetValue(&cmd.path)
+	clause.Arg("repo-path", "Path to the new repository.").Required().PlaceHolder(repoPathPlaceHolder).SetValue(&cmd.path)
```
jpcoenen: Not a full sentence, so there should not be an ending `.`.

internals/secrethub/path_placeholders.go
SimonBarendse: How about just using `secrethub ls [<path>]`?

jpcoenen: Sounds a lot better :+1:

PR #212, opened 2019-10-31
Add --multiline flag to write command

jpcoenen:
If `secrethub write --multiline` or `secrethub write -m` is used, the CLI will keep on asking for input until an `EOF` is reached. This can be triggered by the user pressing `CTRL+D` (which is the same). 

```
$ secrethub write --multiline org/repo/secret
Please type in the value of the secret, followed by [CTRL+D]:
user types
a multiline
secret
```

This flag is not allowed together with `--in-file` or `--clip`.

### DIFF
--- a/internals/cli/ui/ask.go
+++ b/internals/cli/ui/ask.go
@@ -3,6 +3,7 @@ package ui
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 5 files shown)


internals/cli/ui/ask.go
```diff
@@ -57,6 +58,26 @@ func AskSecret(io IO, question string) (string, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = fmt.Fprintf(promptOut, "%s\n", question)
+	if err != nil {
+		return nil, err
+	}
+
+	raw, err := ioutil.ReadAll(promptIn)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

florisvdg:
Works fine 👍 

Couple of things I noticed that are probably okay, but good to be aware of:
- It trims leading and trailing newlines
- The regular secret write doesn't show the output, but the multiline does show the secret in plaintext
- No support for cursors, you get a `^[[D` in the terminal (but not in the actual secret)

PR #216, opened 2019-11-20
Backup Codes

jpcoenen:
Implements the following two commands:
- `secrethub credential backup`: generates a backup code that can be used to recover access to an account when the user does not have another credential.
- `secrethub init`: initializes SecretHub for use on the current device. It gives the option between signing up for a new account and restoring an account from a backup code.

Backup codes are currently in the format of a 64 character hexadecimal string. This will later be replaced by a series of easy-to-type words.


internals/secrethub/init.go
```diff
@@ -0,0 +1,178 @@
...
+		newClientWithoutCredentials: newClientWithoutCredentials,
+		credentialStore:             credentialStore,
+		progressPrinter:             progress.NewPrinter(io.Stdout(), 500*time.Millisecond),
+	}
+}
+
+// Register registers the command, arguments and flags on the provided Registerer.
+func (cmd *InitCommand) Register(r command.Registerer) {
+	clause := r.Command("init", "Initialize the SecretHub client for first use on this device.")
+	clause.Flag("backup-code", "The backup code used for initializing the account on this device.").StringVar(&cmd.backupCode)
```
SimonBarendse: Maybe add something with `restore` or `existing` to make extra clear when this is flag is used.

e.g. `Backup code used to restore an existing account to this device.`

internals/secrethub/init.go
```diff
@@ -0,0 +1,178 @@
...
+
+	if mode == 0 {
+		if cmd.force {
+			return ErrMissingFlags
+		}
+		fmt.Fprintf(cmd.io.Stdout(), "How do you want to initiliaze your SecretHub account on this device?\n"+
+			"1) Signup for a new account\n"+
+			"2) Use a backup code to recover an existing account\n")
+
+		option, err := ui.Ask(cmd.io, "Choose an option [1,2]: ")
```
SimonBarendse: Shall we use `AskAndValidate` to re-prompt when something other than 1 or 2 is chosen?

Or maybe add a `ui.Choose` that does this for you (and lists the options).

internals/secrethub/init.go
jpcoenen: Done.

internals/secrethub/init.go
jpcoenen: :heavy_check_mark: 

Done

PR #217, opened 2019-11-20
Add --parents flag to mkdir to recursively create directories

### LINKED ISSUE #36: Add a --parents flag to the mkdir command
When creating a full tree for a new project, it is often needed to create nested directories. To be able to do so without having to do a lot of monkey work, let's add a `--parents` flag to the `mkdir` command that creates the parent directories as well when these do not exist yet.

This flag would work exactly as the `--parents` flag on the [unix `mkdir`](http://manpages.ubuntu.com/manpages/bionic/man1/mkdir.1.html) command works; It will create the parent directories as well when these do not exist yet and it will not error when all directories already exist.

```
-p, --parents (bool)

   make parent directories as needed
```

SimonBarendse:
It's similar to the unix mkdir --parents flag. It creates all directories in the path that do not exist yet and it does not error when the directory already exists.

Resolves #36


internals/secrethub/mkdir.go
```diff
@@ -33,6 +34,7 @@ func NewMkDirCommand(io ui.IO, newClient newClientFunc) *MkDirCommand {
 func (cmd *MkDirCommand) Register(r command.Registerer) {
 	clause := r.Command("mkdir", "Create a new directory.")
 	clause.Arg("dir-path", "The path to the directory").Required().PlaceHolder(dirPathPlaceHolder).SetValue(&cmd.path)
+	clause.Flag("parents", "Create parent directories if needed. Does not error when directories already exist.").Short('p').BoolVar(&cmd.parents)
```
SimonBarendse: Turns out the `p` shorthand is already in use for `--credential-passphrase`.

jpcoenen: Ugh. I don't like that. Do we even need a shorthand for that? It's dubious flag anyway.

SimonBarendse: Yes, I agree.

How shall we migrate? Are we oke with removing the shorthand or do we want to keep it for backwards compatibility? I don't expect (a lot of) usage and the migration path is really easy (just use `--credential-passphrase` instead), but it can be pretty destructive.

jpcoenen: We can merge it for now without the shorthand, hide the `-p` for credential passphrase and create a deprecation notice for the current short-hand. Then we can add the short-hand to this command later.

SimonBarendse: See #226 for hiding the `-p` shorthand of `--credential-passphrase`

PR #218, opened 2019-11-21
Add credential listing

jpcoenen:
Status is hard-coded to active for now. Will be changed when disabling of credentials is implemented.


internals/secrethub/credential_list.go
```diff
@@ -0,0 +1,76 @@
...
+
+	credentials, err := client.Credentials().ListMine()
+	if err != nil {
+		return err
+	}
+
+	timeFormatter := NewTimeFormatter(cmd.useTimestamps)
+
+	w := tabwriter.NewWriter(cmd.io.Stdout(), 0, 2, 2, ' ', 0)
+	fmt.Fprintln(w,
```
florisvdg: Could be useful to have this in a helper function that just renders a table in a consistent way.

SimonBarendse: If we pick this up, also check out the audit and service ls tables which take an object (`api.AuditEvent`, `api.Service`) and turn it into a table row. This is used to differentiate the table output (e.g. for secret/repo audit events and for key/aws services).
The table writer solution we create should work together with a table row creator solution.

Shall we pick this up in a separate PR? We use the tables in a lot of places, so that would be a bit much to pick up here.

internals/secrethub/credential_list.go
```diff
@@ -0,0 +1,76 @@
...
+		"NAME\t"+
```
SimonBarendse: Currently, this field is always set to `default credential` (which is the server default when no name is supplied). There's no way (yet) to supply this as a user.

Do we want to release listing the name before there's a way to set it?

See this related thread https://github.com/secrethub/secrethub-go/pull/150 on whether to have a name, description or device field.

internals/secrethub/credential_list.go
jpcoenen: Agreed and agreed.

internals/secrethub/credential_list.go
jpcoenen: [`secrethub init` gives the option to supply a name.](https://github.com/secrethub/secrethub-cli/pull/219)

jpcoenen:
Still have to migrate to using Description instead of nam,. And replace  status with `anbled`.

PR #220, opened 2019-11-26
Implement secrethub credential disable command

jpcoenen:


internals/secrethub/credential_disable.go
```diff
@@ -0,0 +1,86 @@
...
+		newClient: newClient,
+	}
+}
+
+// Register registers the command, arguments and flags on the provided Registerer.
+func (cmd *CredentialDisableCommand) Register(r command.Registerer) {
+	clause := r.Command("disable", "Disable a credential for usage on SecretHub.")
+
+	fingerprintHelp := fmt.Sprintf("Fingerprint of the credential to disable. At least the first %d characters must be entered.", api.ShortCredentialFingerprintMinimumLength)
+	clause.Arg("fingerprint", fingerprintHelp).StringVar(&cmd.fingerprint)
```
SimonBarendse: Do we in the future want to be able to select a credential on something other than its fingerprint? (e.g. name, device name)

If we might want to do so, do we want to pass the fingerprint as an arg or a flag?

jpcoenen: Names are not unique. So it would not be possible to use that as a selector through a flag. We _could_ let a user select a credential from a list when we prompt for it. But that is compatible with the current implementation.

PR #224, opened 2019-12-02
Release v0.33.0

jpcoenen:
### Added
- `credential backup`
- `credential disable`
- `credential ls`
- `--parents` flag for `mkdir`, to recursively create all directories in the given path

### Deprecated
- `-p` shorthand for `--credential-passphrase`. Use the full `--credential-passphrase` flag instead.


jpcoenen:
Let's first see whether we want to include #217

PR #234, opened 2019-12-28
Prompt for missing template vars

Marton6:


SimonBarendse:
> Template variables are either set using environment variables: SECRETHUB_VAR_<KEY>=<VALUE> or passed using --var <key>=<value> on the command line.

Both of these are parsed and passed to `RunCommand.templateVars` and `InjectCommand.templateVars`

SimonBarendse:
When adding this functionality, we should do so for both the `run` and the `inject` command.

Marton6:
I will implement this functionality in another pull request.

PR #235, opened 2020-01-06
Prompt missing template variables

Marton6:


internals/secrethub/tpl/v1.go
```diff
@@ -40,8 +40,8 @@ func (p parserV1) Parse(raw string, _, _ int) (Template, error) {
 
 // InjectVars takes a map of template variables with their corresponding values. It replaces
 // the template variables with their values in the template.
-func (t templateV1) Evaluate(vars map[string]string, sr SecretReader) (string, error) {
-	if len(vars) > 0 {
+func (t templateV1) Evaluate(varReader VariableReader, sr SecretReader) (string, error) {
+	if varReader != nil {
```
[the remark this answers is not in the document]
SimonBarendse: Maybe take this check out of the template and use `tpl.IsV1Template` in the commands?

internals/secrethub/run.go
```diff
@@ -147,6 +147,8 @@ func (cmd *RunCommand) Run() error {
 		}
 	}
 
+	variableReader := newVariableReader(templateVars)
```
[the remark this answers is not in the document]
Marton6: Yes, good point!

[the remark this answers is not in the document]
Marton6: Yes

internals/secrethub/variable_reader.go
```diff
@@ -0,0 +1,77 @@
+package secrethub
```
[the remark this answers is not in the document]
Marton6: Sure!

internals/secrethub/variable_reader_test.go
```diff
@@ -0,0 +1,144 @@
...
+		"test4": "testD",
+	}
+
+	reader, err := newVariableReader(osEnv, commandTemplateVars)
+	assert.OK(t, err)
+	io := ui.NewFakeIO()
+	io.PromptIn.Reads = []string{"foobar\n"}
+	reader = newPromptMissingVariableReader(reader, io)
+
+	val, err := reader.ReadVariable("test5")
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

internals/secrethub/variable_reader_test.go
```diff
@@ -0,0 +1,144 @@
...
+		"test3": "testC",
+		"test4": "testD",
+	}
+
+	reader, err := newVariableReader(osEnv, commandTemplateVars)
+	assert.OK(t, err)
+	io := ui.NewFakeIO()
+	reader = newPromptMissingVariableReader(reader, io)
+
+	val, err := reader.ReadVariable("test4")
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

internals/secrethub/variable_reader_test.go
SimonBarendse: err assertion is missing:
```Go
assert.Equal(t, err, nil)
```

internals/secrethub/variable_reader_test.go
SimonBarendse: err assertion is missing:
```Go
assert.Equal(t, err, nil)
```

internals/secrethub/variable_reader_test.go
```diff
@@ -0,0 +1,144 @@
...
+	assert.OK(t, err)
+	io := ui.NewFakeIO()
+	io.PromptIn.Reads = []string{"foobar\n"}
+	reader = newPromptMissingVariableReader(reader, io)
+
+	val, err := reader.ReadVariable("test5")
+	assert.Equal(t, val, "foobar")
+}
+
+func TestPromptVariableReader_NoPrompt(t *testing.T) {
```
SimonBarendse: You could combine these two prompt variable reader testcases to share their setup (as it is the same):

```Go
func TestPromptVariableReader(t *testing.T) {

  // setup code

  t.Run("prompt", func(t *testing.T) {
    // prompt testcase
  })

  t.Run("no prompt", func(t *testing.T) {
    // no prompt testcase
  })
}
```

internals/secrethub/variable_reader_test.go
```diff
@@ -0,0 +1,144 @@
...
+import (
+	"testing"
+
+	"github.com/secrethub/secrethub-cli/internals/cli/ui"
+
+	"github.com/secrethub/secrethub-cli/internals/secrethub/tpl"
+	"github.com/secrethub/secrethub-go/internals/assert"
+)
+
+func TestVariableReader(t *testing.T) {
```
SimonBarendse: Maybe add a testcase where is asserted that variables supplied for the specific command have preference over the variables supplied via the `SECRETHUB_VAR_<KEY>` envvar shared between `run` and `inject`.

This is a case where a key is supplied in both `osEnv` and `commandTemplateVars`. The value from `commandTemplateVars` should be used.

SimonBarendse: Nice! Good to have this tested :ok_hand:

internals/secrethub/variable_reader_test.go
Marton6: Good idea!

internals/secrethub/run_test.go
```diff
@@ -531,6 +536,7 @@ func TestRunCommand_Run(t *testing.T) {
 		},
 		"invalid template var: illegal character": {
```
Marton6: The variable reader is only used when there is an env file provided, thus the validation and the parsing of template variables is only done if `RunCommand.envFile != ""`, as opposed to the current release of the cli, which errors for invalid template variables even if `envFile == ""`. Do you think this is a problem, @SimonBarendse ?

internals/secrethub/tpl/v1.go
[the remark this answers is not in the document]
Marton6: We discussed this in person. The check will be removed entirely.

internals/secrethub/run_test.go
Marton6: Discussed this in person. Not a problem.

internals/secrethub/variable_reader.go
```diff
@@ -0,0 +1,77 @@
...
+		io:     io,
+	}
+}
+
+// ReadVariable fetches a template variable and prompts the user if it is not found.
+func (p *promptMissingVariableReader) ReadVariable(name string) (string, error) {
+	variable, err := p.reader.ReadVariable(name)
+	if err == tpl.ErrTemplateVarNotFound(name) {
+		question := fmt.Sprintf("What is the value of the \"%s\" template variable?", name)
+		return ui.Ask(p.io, question)
```
SimonBarendse: If we're not running locally, we usually cannot ask, because we are running in a headless environment where we cannot prompt.
In those cases, the important thing to return to the user is that the template variable is not found, so that they can supply it using a flag or envvar. Therefore, I think it would be better to return the original template var not found error than the ask error.

Marton6: I did not consider this possibility. Will return the original error.

internals/secrethub/variable_reader.go
```diff
@@ -0,0 +1,81 @@
...
+// ReadVariable fetches a template variable and prompts the user if it is not found.
+func (p *promptMissingVariableReader) ReadVariable(name string) (string, error) {
+	variable, err := p.reader.ReadVariable(name)
+	if err == tpl.ErrTemplateVarNotFound(name) {
+		question := fmt.Sprintf("What is the value of the \"%s\" template variable?\n", name)
+		answer, err := ui.Ask(p.io, question)
+		if err != nil {
+			return "", tpl.ErrTemplateVarNotFound(name)
+		}
+		return answer, nil
```
SimonBarendse: Currently, when the same variable is used multiple times in a template, you are prompted multiple times.

Instead, the template variable value should be remembered and used for all occurrences of the template variable.

Marton6: Good point!

PR #238, opened 2020-01-16
Rephrase template processing error

Marton6:
The CLI will return errors similar to these, when template parsing fails for any reason:

Encountered an error: error while parsing template file 'e1.env': could not parse template at line 1: template is not formatted as key=value pairs (run.invalid_template)  (run.template_file_parsing_failed) 

Encountered an error: error while parsing template file 'e2.env': no value was supplied for template variable 'env' (template.template_var_not_found)  (run.template_file_parsing_failed) 

Encountered an error: error while parsing template file 'e3.env': Secret not found (server.secret_not_found)  (run.template_file_parsing_failed) 

Encountered an error: error while parsing template file 'e4.env': template syntax error at 1:9: expected the closing of a secret tag `}}`, but reached the end of the template. (template.secret_tag_not_closed)  (run.template_file_parsing_failed)

### DIFF
--- a/internals/secrethub/run.go
+++ b/internals/secrethub/run.go
@@ -37,7 +37,7 @@ var (
 	ErrReadEnvFile            = errRun.Code("env_file_read_error").ErrorPref("could not read the environment file %s: %s")
 	ErrEnvDirNotFound         = errRun.Code("env_dir_not_found").Error(fmt.Sprintf("could not find specified environment. Make sure you have executed `%s set`.", ApplicationName))
 	ErrTemplate               = errRun.Code("invalid_template").ErrorPref("could not parse template at line %d: %s")
-	ErrParsingTemplate        = errRun.Code("template_parsing_failed").ErrorPref("error while parsing template file '%s': %s")
+	ErrParsingTemplate        = errRun.Code("template_parsing_failed").ErrorPref("error while processing template file '%s': %s")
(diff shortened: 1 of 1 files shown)


internals/secrethub/run.go
```diff
@@ -37,7 +37,7 @@ var (
 	ErrReadEnvFile            = errRun.Code("env_file_read_error").ErrorPref("could not read the environment file %s: %s")
 	ErrEnvDirNotFound         = errRun.Code("env_dir_not_found").Error(fmt.Sprintf("could not find specified environment. Make sure you have executed `%s set`.", ApplicationName))
 	ErrTemplate               = errRun.Code("invalid_template").ErrorPref("could not parse template at line %d: %s")
-	ErrTemplateFile           = errRun.Code("invalid_template_file").ErrorPref("template file '%s' is invalid: %s")
+	ErrParsingTemplate        = errRun.Code("template_parsing_failed").ErrorPref("error while parsing template file '%s': %s")
```
SimonBarendse: We could also make it "error while *processing* template file:", as it seems the error might come from various phases of processing, including, but not limited to parsing.
It could come for example, from looking up the secret mentioned in the file, from reading or parsing the credential to authenticate to the SecretHub API server to do do so etc.

Marton6: Good point!

PR #239, opened 2020-01-16
Add custom error when secret is not found

Marton6:


Marton6:
> The path is really nice feedback to have when a secret cannot be found, so that you can better determine where the problem is.
> 
> detective gif
> 
> I think this is valuable for a lot more usecases, for example:
> 
> * `secrethub inject`
> * (multiple) `secrethub read`s, e.g. in a script: `./my-script.sh`
> * (multiple ) reads by the Terraform provider
> 
> Therefore, shall we add this in [secrethub-go](https://github.com/secrethub/secrethub-go) instead, so that every secret not found error includes the path in the error message?

Good point! I will implement it in the client.

PR #243, opened 2020-02-14
Substitute os env vars with the 'secrethub://' prefix

Marton6:


internals/secrethub/run.go
```diff
@@ -422,6 +422,48 @@ func ReadEnvFile(filepath string, varReader tpl.VariableReader, parser tpl.Parse
 	}, nil
 }
 
+// OsEnv is an environment with secrets configured with the
+// secrethub:// syntax in the os environment variables.
+type OsEnv struct {
```
Marton6: Any ideas for a better name?

internals/secrethub/run.go
```diff
@@ -422,6 +422,48 @@ func ReadEnvFile(filepath string, varReader tpl.VariableReader, parser tpl.Parse
...
+		envVars: envVars,
+	}
+}
+
+// Env returns a map of key value pairs with the secrets configured with the
+// secrethub:// syntax.
+func (env *OsEnv) Env(secrets map[string]string, _ tpl.SecretReader) (map[string]string, error) {
+	envVarsWithSecrets := make(map[string]string)
+	for key, path := range env.envVars {
+		envVarsWithSecrets[key] = secrets[path]
```
Marton6: Should I use the `SecretReader` or the secrets from the map?

Marton6: Right now, with the `--ignore-missing-secrets` flag the environment variables with missing secrets in them are cleared. Is this the expected behavior?

SimonBarendse: > Should I use the SecretReader or the secrets from the map?

I believe you can currently choose; All secret paths returned with `Secrets` are present in the `secrets` map. So you can either return the secret paths in `Secrets` and use the `secrets` map, or return an empty list with `Secrets` and use the secret reader.

I think this was introduced together with template variables. For template variables we can't determine in advance what the secret paths will be (as they dependent on evaluating the template variables). That's when the secret reader was introduced so that the secrets can be read when evaluating the template.

It would be nice if we refactor all sources to use the secret reader, so that we can remove the need for the `secrets` parameter and remove this unclear choice of which to use. Then we can change the signature to `Env(sr tpl.SecretReader)` and we no longer have this un-clarity. We can pick that up as a separate task/PR.

SimonBarendse: > Right now, with the --ignore-missing-secrets flag the environment variables with missing secrets in them are cleared. Is this the expected behavior?

What does `are cleared` mean? Are the environment variables set with an empty value or not set at all?

The help text says:
```
      --ignore-missing-secrets   Do not return an error when a secret does not
                                 exist and use an empty value instead.
```

So I'd expect the secret reference `{{ path/to/secret }}` to be replaced with an empty string. I would expect the environment variable to be set.

internals/secrethub/run.go
SimonBarendse: This is a though one for naming. Maybe something with reference? As environment variable contains a secret reference. e.g. `referenceEnv`.

Whatever the name, please make it private. Then we can always change the name later. When making this functionality public to other packages (e.g. to be used from end-user Golang applications), we should move it to its own package.

internals/secrethub/run.go
```diff
@@ -422,6 +424,48 @@ func ReadEnvFile(filepath string, varReader tpl.VariableReader, parser tpl.Parse
...
+	envVars map[string]string
+}
+
+// NewOsEnvSource returns an environment with secrets configured in the
+// os environment with the secrethub:// syntax.
+func NewOsEnvSource(osEnv map[string]string) *OsEnv {
+	envVars := make(map[string]string)
+	for key, value := range osEnv {
+		if strings.HasPrefix(value, "secrethub://") {
+			envVars[key] = strings.TrimPrefix(value, "secrethub://")
```
jpcoenen: I'd move the `secrethub://` value to constant. Duplicate constants are evil :grimacing:

internals/secrethub/run.go
jpcoenen: What about `SecretReferenceEnv`?

internals/secrethub/run_test.go
```diff
@@ -544,6 +550,46 @@ func TestRunCommand_Run(t *testing.T) {
...
+									Err: api.ErrSecretNotFound,
+								},
+							},
+						},
+					}, nil
+				},
+			},
+			err: api.ErrSecretNotFound,
+		},
+		"os env secret not found ignored": {
```
jpcoenen: :ok_hand:

internals/secrethub/run.go
jpcoenen: `referenceEnv` might be better.

jpcoenen: `referenceEnv` might be better. And I agree on it being unexported.

internals/secrethub/run.go
```diff
@@ -422,6 +424,44 @@ func ReadEnvFile(filepath string, varReader tpl.VariableReader, parser tpl.Parse
...
 
+// referenceEnv is an environment with secrets configured with the
+// secrethub:// syntax in the os environment variables.
+type referenceEnv struct {
+	envVars map[string]string
+}
+
+// NewReferenceEnv returns an environment with secrets configured in the
+// os environment with the secrethub:// syntax.
+func NewReferenceEnv(osEnv map[string]string) *referenceEnv {
```
golangcibot: exported func NewReferenceEnv returns unexported type *github.com/secrethub/secrethub-cli/internals/secrethub.referenceEnv, which can be annoying to use (from `golint`)

internals/secrethub/run.go
Marton6: `referenceEnv` sounds good :+1: 
Made it private.

internals/secrethub/run.go
Marton6: Thanks for the explanation of sources, I modified the implementation to use the `tpl.SecretReader`.

The variables are set to an empty value.

internals/secrethub/run.go
Marton6: Sure thing, fixed it!

internals/secrethub/inject.go
```diff
@@ -34,6 +34,7 @@ type InjectCommand struct {
 	useClipboard                  bool
 	clearClipboardAfter           time.Duration
 	clipper                       clip.Clipper
+	osEnv                         func() []string
```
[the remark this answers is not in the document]
Marton6: You're right, it can just be a `[]string` :sweat_smile:

jpcoenen:
Code-wise :+1: 

@florisvdg did you do a functional review?

PR #248, opened 2020-02-21
Rename deprecated nfpms.name_template field

SimonBarendse:
The old name_template is deprecated in favor of new file_name_template
field.

See: https://goreleaser.com/deprecations/#nfpms-name-template


SimonBarendse:
@florisvdg Could you review?

PR #250, opened 2020-02-21
Refactor run command

Marton6:


internals/secrethub/run.go
```diff
@@ -112,66 +116,156 @@ func (cmd *RunCommand) Register(r command.Registerer) {
...
 		}
 
 		parser, err := getTemplateParser(raw, cmd.templateVersion)
 		if err != nil {
-			return err
+			return nil, nil, err
 		}
 
-		envFile, err := ReadEnvFile(cmd.envFile, templateVariableReader, parser)
+		envFile, err := ReadEnvFile(cmd.envFile, bytes.NewReader(raw), templateVariableReader, parser)
```
Marton6: Modified `ReadEnvFile` to take the file name and a reader to its contents, instead of opening the env file again.

internals/secrethub/run.go
```diff
@@ -667,13 +676,6 @@ type EnvDir map[string]string
 // NewEnvDir sources environment variables from files in a given directory,
 // using the file name as key and contents as value.
 func NewEnvDir(path string) (EnvDir, error) {
-	_, err := os.Stat(path)
-	if os.IsNotExist(err) {
-		return nil, ErrEnvDirNotFound
-	} else if err != nil {
-		return nil, ErrReadEnvDir(err)
-	}
-
```
Marton6: Removed these checks because `NewEnvDir` is called like this:
```go
_, err = cmd.osStat(envDir)
if err == nil {
	dirSource, err := NewEnvDir(envDir)
	if err != nil {
		return nil, nil, err
	}
	envSources = append(envSources, dirSource)
}
```

So the error will always be nil.

PR #253, opened 2020-02-27
Fix order of env sources

Marton6:


SimonBarendse:
To give some context, the order of precedence is defined as follows:

1. Environment variables configured on the command-line with the `--envar KEY=PATH` flag, e.g. `--envar DB_PASSWORD=path/to/secret`
 1. Environment variables configured with the secret reference syntax in environment variables, e.g. `secrethub://path/to/secret`
1. Environment variables configured in a `.env` file. By default `secrethub.env` is used; Use the `--env-file` flag to use a different file.
 1. Environment variables from the parent environment.

mackenbach:
> To give some context, the order of precedence is defined as follows:
> 
> 1. Environment variables configured on the command-line with the `--envar KEY=PATH` flag, e.g. `--envar DB_PASSWORD=path/to/secret`
> 2. Environment variables configured with the secret reference syntax in environment variables, e.g. `secrethub://path/to/secret`
> 3. Environment variables configured in a `.env` file. By default `secrethub.env` is used; Use the `--env-file` flag to use a different file.
> 4. Environment variables from the parent environment.

Is this documented somewhere? If not, it should at least be documented on the website.

SimonBarendse:
@mackenbach This bug was discovered before the secrethub reference syntax docs were published. Therefore, we chose to release this bugfix first, before publishing the docs. We will publish these today.

PR #255, opened 2020-02-28
Add `env read` and `env ls` commands

Marton6:


internals/secrethub/env_source.go
```diff
@@ -0,0 +1,563 @@
...
+			return nil, err
+		}
+	}
+
+	return flags, nil
+}
+
+// Env returns a map of environment variables sourced from
+// command-line flags and set to their corresponding value.
+func (ef EnvFlags) Env() (map[string]value, error) {
```
golangcibot: exported method Env returns unexported type map[string]github.com/secrethub/secrethub-cli/internals/secrethub.value, which can be annoying to use (from `golint`)

internals/secrethub/env_source.go
```diff
@@ -0,0 +1,563 @@
...
+
+			env[f.Name()] = newEnvDirSecretValue(string(fileContent))
+		}
+	}
+
+	return env, nil
+}
+
+// Env returns a map of environment variables sourced from the directory.
+func (dir EnvDir) Env() (map[string]value, error) {
```
golangcibot: exported method Env returns unexported type map[string]github.com/secrethub/secrethub-cli/internals/secrethub.value, which can be annoying to use (from `golint`)

internals/secrethub/env_source.go
```diff
@@ -0,0 +1,563 @@
...
+}
+
+// EnvFile contains an environment that is read from a file.
+type EnvFile struct {
+	path string
+	env  EnvSource
+}
+
+// Env returns a map of key value pairs read from the environment file.
+func (e EnvFile) Env() (map[string]value, error) {
```
golangcibot: exported method Env returns unexported type map[string]github.com/secrethub/secrethub-cli/internals/secrethub.value, which can be annoying to use (from `golint`)

internals/secrethub/env_source.go
```diff
@@ -0,0 +1,563 @@
...
+
+type plaintextValue struct {
+	value string
+}
+
+func newPlaintextValue(value string) *plaintextValue {
+	return &plaintextValue{value: value}
+}
+
+func (v *plaintextValue) resolve(_ tpl.SecretReader) (string, error) {
```
golangcibot: (*plaintextValue).resolve - result 1 (error) is always nil (from `unparam`)

internals/secrethub/env_source.go
```diff
@@ -0,0 +1,563 @@
...
+
+	// --envar flag
+	// TODO: Validate the flags when parsing by implementing the Flag interface for EnvFlags.
+	flagEnv, err := NewEnvFlags(env.envar)
+	if err != nil {
+		return nil, err
+	}
+	sources = append(sources, flagEnv)
+
+	var envs []map[string]value
```
golangcibot: Consider preallocating `envs` (from `prealloc`)

internals/secrethub/env_source.go
```diff
@@ -0,0 +1,563 @@
...
+
+func (v *plaintextValue) resolve(_ tpl.SecretReader) (string, error) {
+	return v.value, nil
+}
+
+func (v *plaintextValue) containsSecret() bool {
+	return false
+}
+
+type osEnv struct {
```
golangcibot: type `osEnv` is unused (from `unused`)

internals/secrethub/env_source.go
```diff
@@ -0,0 +1,563 @@
...
+			key:        key,
+			value:      value,
+			lineNumber: -1,
+		}
+		i++
+	}
+	return vars, nil
+}
+
+type plaintextValue struct {
```
golangcibot: type `plaintextValue` is unused (from `unused`)

internals/secrethub/env_source.go
```diff
@@ -0,0 +1,563 @@
...
+	value string
+}
+
+func newPlaintextValue(value string) *plaintextValue {
```
golangcibot: func `newPlaintextValue` is unused (from `unused`)

internals/secrethub/run.go
```diff
@@ -212,114 +189,48 @@ func (cmd *RunCommand) Run() error {
...
+	newEnv := map[string]string{}
+	for name, value := range osEnv {
+		newEnv[name] = value
 	}
 
+	// add values from .secretsenv and other sources
+	var dirValues map[string]value
 	envDir := filepath.Join(secretspec.SecretEnvPath, cmd.env)
-	_, err = cmd.osStat(envDir)
+	_, err := os.Stat(envDir)
```
Marton6: This needs to be changed to allow for injecting `osStat`.

internals/secrethub/run_test.go
```diff
@@ -544,32 +560,38 @@ func TestRunCommand_Run(t *testing.T) {
...
 		},
 		"os env secret not found": {
 			command: RunCommand{
-				osEnv:   []string{"TEST=secrethub://nonexistent/secret/path"},
 				command: []string{"echo", "test"},
 				io:      ui.NewFakeIO(),
-				osStat:  osStatNotExist,
+				environment: &environment{
+					osEnv:  []string{"TEST=secrethub://nonexistent/secret/path"},
+					osStat: osStatNotExist,
```
Marton6: With the current implementation `osStat` and `osEnv` should also be injected into the `RunCommand`.

internals/secrethub/env_ls.go
```diff
@@ -0,0 +1,50 @@
...
+func (cmd *EnvListCommand) Register(r command.Registerer) {
+	clause := r.Command("ls", "Read the value of a single environment variable.")
+	clause.Alias("list")
+
+	cmd.environment.register(clause)
+
+	command.BindAction(clause, cmd.Run)
+}
+
+// Run handles the command with the options as specified in the command.
```
mackenbach: Super tiny thing and not required to fix (now), but saw your PR and was curious and scrolled past this piece of documentation. Consider putting something a bit simpler here as this sentence contains a few abstract concepts that can add more noise than they explain.

What is the key thing you need to understand about this function? Consider:

```go
// Run executes the command. This is where the actual logic of the command lives.
```
Or even simpler:
```go
// Run executes the command. 
```

SimonBarendse:
I've used this functionality successfully in this GitHub Action: https://github.com/SimonBarendse/github-actions-test/runs/474988530

internals/secrethub/env_ls.go
```diff
@@ -0,0 +1,50 @@
...
+func NewEnvListCommand(io ui.IO) *EnvListCommand {
+	return &EnvListCommand{
+		io:          io,
+		environment: newEnvironment(io),
+	}
+}
+
+// Register adds a CommandClause and it's args and flags to a Registerer.
+func (cmd *EnvListCommand) Register(r command.Registerer) {
+	clause := r.Command("ls", "Read the value of a single environment variable.")
```
florisvdg: This is not correct.

jpcoenen:
I have just taken a quick look at this and my first reaction is: "why is this in the CLI?" I understand that it is necessary to get GitHub Actions to work, but it kind of feels like these commands are tailored specifically to GitHub actions. 

For example, something like the `env ls` command can be really powerful for the user if he wants to debug. If it were to give a table with: 
- environment variable name
- source (`.secrethub.env`, `reference`, `flag`)
- value (secrets not yet replaced unless `--reveal-secrets` is used)

In that case, the current feature would be a small subset of the command. For example `secrethub env ls --names-only`. If we already do it that way now, we won't run into compatibility-issues later.

What do you guys think?

mackenbach:
Yes, you're right @jpcoenen. It is definitely a feature that has been driven by a very specific use case and we haven't designed it for other use cases. So purely based on that, you could make a strong argument that this shouldn't be in the CLI per se.

Reason this happened was because we are on a tight schedule to deliver a GitHub Action. We could take a moment to design this now and implement it, or we could relax our forward compatibility constraints here given that the command is still 'hidden' and in 'beta'.

If we leave it as is for now, we should definitely think about the update strategy for the action and if people pin to this first version, whether that version will break (e.g. when it downloads a newer backwards incompatible version of the CLI?)

florisvdg:
We discussed that for real debugging we could create `secrethub env trace`, which will show you a graph of where the secret comes from, just like dependency managers have to see how certain  dependencies and versions got included and resolved.

SimonBarendse:
> It is definitely a feature that has been driven by a very specific use case and we haven't designed it for other use cases.

As a matter of fact, Floris and I did discuss other use cases, for which we came up with a.o. the `env trace` command. Might be good to walk you guys (@jpcoenen @mackenbach) through our reasoning and ideas for future improvements/additions, so that we can review this PR in the context of that bigger picture.

> If we leave it as is for now, we should definitely think about the update strategy for the action and if people pin to this first version, whether that version will break (e.g. when it downloads a newer backwards incompatible version of the CLI?)

The functionality is currently hidden. Our idea is to release it as a beta feature, for which we don't guarantee backwards compatibility. We'll communicate that clearly in the release notes.

jpcoenen:
> The functionality is currently hidden. Our idea is to release it as a beta feature, for which we don't guarantee backwards compatibility. We'll communicate that clearly in the release notes.

Okay, if we change it, what happens with existing GitHub Actions?

SimonBarendse:
>> The functionality is currently hidden. Our idea is to release it as a beta feature, for which we don't guarantee backwards compatibility. We'll communicate that clearly in the release notes.
>
> Okay, if we change it, what happens with existing GitHub Actions?

Good question. I imagine GitHub Actions being in beta for at least as long as the `env ls` and `env read` commands. When we're making GitHub Actions GA, we should also take the `env` commands out of beta.

florisvdg:
We can just fix the CLI version number in the Action if we want. That does mean that there's an extra step again when releasing a new version of the CLI.

jpcoenen:
Are Actions themselves versioned? Or do people always use the latest?

SimonBarendse:
> Are Actions themselves versioned? Or do people always use the latest?

_"Workflows can reference specific versions of actions using a commit SHA, branch, or tag."_

https://help.github.com/en/actions/building-actions/about-actions#versioning-your-action

florisvdg:
Even if people choose to use the latest Action, we can technically do what we want with the CLI and the internals of the Action, as long as the functionality of it stays the same.

SimonBarendse:
Yes good point Floris. If we make the CLI version used by a version of the GitHub Action fixed, we can make the GitHub Action GA even if it depends on beta functionality of the CLI. That just means that when we're breaking the functionality in the CLI, we have to replace it with something else in the GitHub Action. We want to do that anyways, as we do want to keep the GitHub Action.

jpcoenen:
I'm okay with the proposed plan of releasing it hidden. My review request can be dismissed.

internals/secrethub/env_ls.go
```diff
@@ -23,7 +23,8 @@ func NewEnvListCommand(io ui.IO) *EnvListCommand {
 
 // Register adds a CommandClause and it's args and flags to a Registerer.
 func (cmd *EnvListCommand) Register(r command.Registerer) {
-	clause := r.Command("ls", "Read the value of a single environment variable.")
+	clause := r.Command("ls", "[BETA] List environment variable keys that will be injected with secrets.")
```
florisvdg: How about: `List environment variable names that will be populated with secrets`

PR #258, opened 2020-03-05
Refactor tests to use fakeclient with funcs

Marton6:
Updated the tests after refactoring `fakeclient` in:
https://github.com/secrethub/secrethub-go/pull/176

### DIFF
--- a/internals/secrethub/acl_check_test.go
+++ b/internals/secrethub/acl_check_test.go
@@ -109,12 +109,12 @@ func TestACLCheckCommand_Run(t *testing.T) {
 			tc.cmd.io = io
 
 			lister := tc.lister
-
+			var argPath string
 			tc.cmd.newClient = func() (secrethub.ClientInterface, error) {
 				return fakeclient.Client{
 					AccessRuleService: &fakeclient.AccessRuleService{
 						ListLevelsFunc: func(path string) ([]*api.AccessLevel, error) {
-							assert.Equal(t, path, tc.listerArgPath)
+							argPath = path
 							return lister(path)
 						},
 					},
@@ -127,6 +127,7 @@ func TestACLCheckCommand_Run(t *testing.T) {
 			// Assert
 			assert.Equal(t, err, tc.err)
 			assert.Equal(t, io.StdOut.String(), tc.out)
+			assert.Equal(t, argPath, tc.listerArgPath)
 		})
 	}
 }
--- a/internals/secrethub/acl_check_test.go
+++ b/internals/secrethub/acl_check_test.go
@@ -18,7 +18,7 @@ func TestACLCheckCommand_Run(t *testing.T) {
 	cases := map[string]struct {
 		cmd           ACLCheckCommand
 		newClientErr  error
-		lister        fakeclient.AccessLevelLister
+		lister        func(path string) ([]*api.AccessLevel, error)
 		listerArgPath api.Path
 		out           string
 		err           error
@@ -32,8 +32,8 @@ func TestACLCheckCommand_Run(t *testing.T) {
 				accountName: "dev1",
 				path:        "namespace/repo",
 			},
-			lister: fakeclient.AccessLevelLister{
-				ReturnsAccessLevels: []*api.AccessLevel{
+			lister: func(path string) ([]*api.AccessLevel, error) {
+				return []*api.AccessLevel{
 					{
 						Account: &api.Account{
 							Name: "dev1",
@@ -46,7 +46,7 @@ func TestACLCheckCommand_Run(t *testing.T) {
 						},
 						Permission: api.PermissionWrite,
 					},
-				},
+				}, nil
 			},
 			listerArgPath: "namespace/repo",
 			out:           "read\n",
@@ -56,15 +56,15 @@ func TestACLCheckCommand_Run(t *testing.T) {
 				accountName: "dev1",
 				path:        "namespace/repo",
 			},
-			lister: fakeclient.AccessLevelLister{
-				ReturnsAccessLevels: []*api.AccessLevel{
+			lister: func(path string) ([]*api.AccessLevel, error) {
+				return []*api.AccessLevel{
 					{
 						Account: &api.Account{
 							Name: "dev2",
 						},
 						Permission: api.PermissionWrite,
 					},
-				},
+				}, nil
 			},
 			listerArgPath: "namespace/repo",
 			out:           "none\n",
@@ -73,8 +73,8 @@ func TestACLCheckCommand_Run(t *testing.T) {
 			cmd: ACLCheckCommand{
 				path: "namespace/repo",
 			},
-			lister: fakeclient.AccessLevelLister{
-				ReturnsAccessLevels: []*api.AccessLevel{
+			lister: func(path string) ([]*api.AccessLevel, error) {
+				return []*api.AccessLevel{
 					{
 						Account: &api.Account{
 							Name: "dev1",
@@ -87,7 +87,7 @@ func TestACLCheckCommand_Run(t *testing.T) {
 						},
 						Permission: api.PermissionWrite,
 					},
-				},
+				}, nil
 			},
 			listerArgPath: "namespace/repo",
 			out: "PERMISSIONS    ACCOUNT\n" +
@@ -95,8 +95,8 @@ func TestACLCheckCommand_Run(t *testing.T) {
 				"read           dev1\n",
 		},
 		"list error": {
-			lister: fakeclient.AccessLevelLister{
-				Err: testError,
+			lister: func(path string) ([]*api.AccessLevel, error) {
+				return nil, testError
 			},
 			err: testError,
 		},
@@ -108,12 +108,15 @@ func TestACLCheckCommand_Run(t *testing.T) {
 			io := ui.NewFakeIO()
 			tc.cmd.io = io
 
-			lister := &tc.lister
+			lister := tc.lister
 
 			tc.cmd.newClient = func() (secrethub.ClientInterface, error) {
 				return fakeclient.Client{
 					AccessRuleService: &fakeclient.AccessRuleService{
-						LevelLister: lister,
+						ListLevelsFunc: func(path string) ([]*api.AccessLevel, error) {
+							assert.Equal(t, path, tc.listerArgPath)
+							return lister(path)
+						},
 					},
 				}, tc.newClientErr
 			}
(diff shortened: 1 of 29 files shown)


internals/secrethub/acl_check_test.go
```diff
@@ -108,12 +108,15 @@ func TestACLCheckCommand_Run(t *testing.T) {
...
 
-			lister := &tc.lister
+			lister := tc.lister
 
 			tc.cmd.newClient = func() (secrethub.ClientInterface, error) {
 				return fakeclient.Client{
 					AccessRuleService: &fakeclient.AccessRuleService{
-						LevelLister: lister,
+						ListLevelsFunc: func(path string) ([]*api.AccessLevel, error) {
+							assert.Equal(t, path, tc.listerArgPath)
```
SimonBarendse: What happens if `List` is never called? Then this assertion is never made.

I believe we need to either use a struct to store the value or declare a variable in the scope where the assertion was previously made and check there. What do you think?

Marton6: Yes, I accidentally missed that :sweat_smile:
I agree, that seems like a good approach. Fixed it!

internals/secrethub/acl_list_test.go
```diff
@@ -40,17 +40,27 @@ func TestACLListCommand_run(t *testing.T) {
...
 			accessrules: fakeclient.AccessRuleService{
-				Lister: &fakeclient.AccessRuleLister{
-					Err: testError,
+				ListFunc: func(path string, depth int, ancestors bool) ([]*api.AccessRule, error) {
+					return nil, testError
+				},
+			},
+			dirs: fakeclient.DirService{
+				GetTreeFunc: func(path string, depth int, ancestors bool) (*api.Tree, error) {
+					return nil, nil
```
SimonBarendse: Before, there was no function provided to get the tree. Is there a specific reason you changed this? I think the cases are easier to understand if we only set the fields that are used (and leave the other fields to their zero values). What do you think?

SimonBarendse:
Can you update the go client version used so that the CI pipeline can run?

internals/secrethub/acl_list_test.go
Marton6: This was a mistake, fixed it :+1:

SimonBarendse: @Marton6 This was not the only occurrence. There's several additions like this. Could you update the others as well?

Marton6: The others (in this file, at least) are required.

internals/secrethub/acl_list_test.go
```diff
@@ -40,17 +40,22 @@ func TestACLListCommand_run(t *testing.T) {
...
 		"0 access rules": {
 			cmd: ACLListCommand{},
 			accessrules: fakeclient.AccessRuleService{
-				Lister: &fakeclient.AccessRuleLister{
-					ReturnsAccessRules: []*api.AccessRule{},
+				ListFunc: func(path string, depth int, ancestors bool) ([]*api.AccessRule, error) {
+					return []*api.AccessRule{}, nil
+				},
+			},
+			dirs: fakeclient.DirService{
```
SimonBarendse: here

internals/secrethub/acl_list_test.go
```diff
@@ -62,8 +67,13 @@ func TestACLListCommand_run(t *testing.T) {
 				ancestors: true,
 			},
 			accessrules: fakeclient.AccessRuleService{
-				Lister: &fakeclient.AccessRuleLister{
-					ReturnsAccessRules: []*api.AccessRule{},
+				ListFunc: func(path string, depth int, ancestors bool) ([]*api.AccessRule, error) {
+					return []*api.AccessRule{}, nil
+				},
+			},
+			dirs: fakeclient.DirService{
```
SimonBarendse: here

internals/secrethub/org_inspect_test.go
```diff
@@ -107,29 +107,42 @@ func TestOrgInspectCommand_Run(t *testing.T) {
...
 			orgService: fakeclient.OrgService{
-				MemberService: &fakeclient.OrgMemberService{
-					Lister: fakeclient.OrgMemberLister{
-						Err: testErr,
+				MembersService: &fakeclient.OrgMemberService{
+					ListFunc: func(org string) ([]*api.OrgMember, error) {
+						return nil, testErr
 					},
 				},
+				GetFunc: func(name string) (*api.Org, error) {
```
SimonBarendse: here

internals/secrethub/org_inspect_test.go
```diff
@@ -107,29 +107,42 @@ func TestOrgInspectCommand_Run(t *testing.T) {
...
 		},
 		"list repos error": {
 			orgService: fakeclient.OrgService{
-				MemberService: &fakeclient.OrgMemberService{},
+				MembersService: &fakeclient.OrgMemberService{
+					ListFunc: func(org string) ([]*api.OrgMember, error) {
+						return nil, nil
+					},
+				},
+				GetFunc: func(name string) (*api.Org, error) {
```
SimonBarendse: here

internals/secrethub/org_inspect_test.go
```diff
@@ -107,29 +107,42 @@ func TestOrgInspectCommand_Run(t *testing.T) {
...
+					return &api.Org{}, nil
+				},
 			},
 			repoService: fakeclient.RepoService{
-				Lister: fakeclient.RepoLister{
-					Err: testErr,
+				GetFunc: func(path string) (*api.Repo, error) {
```
SimonBarendse: here

internals/secrethub/org_ls_test.go
```diff
@@ -71,14 +71,19 @@ func TestOrgLsCommand_run(t *testing.T) {
...
+					return []*api.Org{
 						{
 							Name: "company1",
 						},
 						{
 							Name: "company2",
 						},
+					}, nil
+				},
+				MembersService: &fakeclient.OrgMemberService{
```
SimonBarendse: here

internals/secrethub/org_ls_test.go
```diff
@@ -91,39 +96,48 @@ func TestOrgLsCommand_run(t *testing.T) {
 		},
 		"orgs mine error": {
 			service: fakeclient.OrgService{
-				MineLister: fakeclient.OrgMineLister{
-					Err: testErr,
+				ListMineFunc: func() ([]*api.Org, error) {
+					return nil, testErr
+				},
+				MembersService: &fakeclient.OrgMemberService{
```
SimonBarendse: here

internals/secrethub/org_ls_test.go
```diff
@@ -91,39 +96,48 @@ func TestOrgLsCommand_run(t *testing.T) {
...
 		"list repos error": {
 			service: fakeclient.OrgService{
-				MineLister: fakeclient.OrgMineLister{
-					ReturnsOrgs: []*api.Org{
+				ListMineFunc: func() ([]*api.Org, error) {
+					return []*api.Org{
 						{},
+					}, nil
+				},
+				MembersService: &fakeclient.OrgMemberService{
```
SimonBarendse: here

internals/secrethub/acl_list_test.go
SimonBarendse: Yes, they are indeed in other files. Have pointed at a view. They're easy to spot if you look at the diff. So please go over the other files there as well.

internals/secrethub/org_ls_test.go
Marton6: Fixed it!

internals/secrethub/org_ls_test.go
Marton6: Fixed it!

internals/secrethub/org_ls_test.go
Marton6: Fixed it!

internals/secrethub/org_inspect_test.go
Marton6: This one is required.

internals/secrethub/org_inspect_test.go
Marton6: This one is required.

internals/secrethub/org_inspect_test.go
Marton6: This one is required.

internals/secrethub/acl_list_test.go
Marton6: This one is required.

internals/secrethub/acl_list_test.go
Marton6: This one is required.

Marton6:
I went through it but I couldn't find any occurrences that are not required.

internals/secrethub/org_inspect_test.go
SimonBarendse: Can you explain why? It wasn't supplied before, right?

Marton6: Sorry, I should have clarified all this when opening the PR, so here is my explanation:
Before the change, we were using structs for mocks, for example the `OrgGetter`: 

```go
type OrgGetter struct {
	ArgName    string
	ReturnsOrg *api.Org
	Err        error
}


// Get saves the arguments it was called with and returns the mocked response.
func (g *OrgGetter) Get(name string) (*api.Org, error) {
	g.ArgName = name
	return g.ReturnsOrg, g.Err
}
```
If the the OrgGetter was not initialized it returned the zero values `(nil, nil)`.

In contrast, when using functions, an uninitialized `func` field has the value of `nil` and panics when `OrgService.Get` is called.
```go
func (s *OrgService) Get(name string) (*api.Org, error) {
	return s.GetFunc(name)
}
```

I thought about the drawback that now we have to initialize more fields when testing, however I realized it might be actually be an advantage as it forces the tester to initialize all used client functions explicitly and makes it more clear what the test is doing (in my opinion).

mackenbach: Yes that's great. It's totally fine for tests to fail/panic if a developer hasn't set all the required fields of the fake. IMO, that's actually a feature, not a bug.

SimonBarendse: > I thought about the drawback that now we have to initialize more fields when testing, however I realized it might be actually be an advantage as it forces the tester to initialize all used client functions explicitly and makes it more clear what the test is doing (in my opinion).

I agree that it's an advantage that we're now explicit about that a function is called and what is returned :+1:

Issue #261, opened 2020-03-18
Add `--force` flag to inject command


SimonBarendse:
```
-f, --force (bool)
  When injecting to an output file and the file already exists, overwrite it.
```

mackenbach:
Sweet! Should this also be included in the `read` command, which has an `--output-file` flag?

And, it's kinda implied, but it may be good to speficy that the `--force` flag is ignored if `--output-file` isn't specified?

mackenbach:
And thanks for the suggestion @antonbabenko

SimonBarendse:
> Sweet! Should this also be included in the read command, which has an --output-file flag?

The read command always overwrites the file. I'm hesitant to change that, as it might break existing workflows if we now require `--force` to do so.

> And, it's kinda implied, but it may be good to speficy that the --force flag is ignored if --output-file isn't specified?

Yes good idea. Can't be too explicit :+1:

mackenbach:
> The read command always overwrites the file. I'm hesitant to change that, as it might break existing workflows if we now require --force to do so.

After thinking twice about what you said, do we actually want that, and why? Yes, changing it would be breaking, but I'd like us to be conscious as to why we have this behavior. 

And the docs currently don't mention that it always overwrites, so maybe it's good to add that there too to be explicit.

```
-o, --out-file=OUT-FILE  Write the secret value to this file.
```

SimonBarendse:
Ideally, the `read` command should behave similar to the `inject` command. Also, I think passing a flag to explicitly overwrite the file is better.

I can see the following migration path:

Add `--force` flag to read to overwrite the file, also, still overwrite if the flag is not supplied as well, but deprecate that. Then, a few versions later, remove support for overwriting the file without passing the `--force` flag.

This way, we give users that depend on the `read` command to overwrite the file some time to migrate.

PR #264, opened 2020-03-19
Update --force flag help text

Marton6:


internals/secrethub/flags.go
```diff
@@ -15,5 +15,5 @@ func registerTimestampFlag(r FlagRegisterer) *kingpin.FlagClause {
 }
 
 func registerForceFlag(r FlagRegisterer) *kingpin.FlagClause {
-	return r.Flag("force", "Ignore confirmation and fail instead of prompt for missing arguments.").Short('f')
+	return r.Flag("force", "Disable all confirmation prompts.").Short('f')
```
SimonBarendse: Please include that this overwrites the output file if it already exists. See this ticket: https://github.com/secrethub/secrethub-cli/issues/261

Marton6: Isn't that specific to the `secrethub inject` command?

SimonBarendse: Yes correct. Should have been more explicit about that. My suggestion is to not use this function, but instead register the same flag with a more explicit message that explains exactly what it means for the `inject` command.

Marton6: Got it, thanks!

SimonBarendse: Please also include Marc's suggestion of making explicit that the `--force` flag is ignored if no `--output-file` is supplied. See the ticket: https://github.com/secrethub/secrethub-cli/issues/261#issuecomment-601125693

Marton6: Done

PR #265, opened 2020-03-20
Fix mixing stdout and stderr

SimonBarendse:
Output should be written in the same order it was received, regardless of the target output stream. Before this commit, it was possible that output on stdout and stderr were mixed, because one of them was delayed.


jpcoenen:
This is on hold until #267 is resolved.

jpcoenen:
Think we can close this for now because of #267.

Issue #196, opened 2019-10-10
stdout and stderr mixed in secrethub run


SimonBarendse:
### Encountered Behavior
When the subcommand of `run` writes to both `stdout` and `stderr` and secret masking is enabled, the output of stdout and stderr is mixed, resulting in potentially unreadable output.

When wrapping [this example](https://goplay.space/#-HWl4aV2FYY) in `secrethub run` the output can be any of these:
`123`
`213`
`132`

### Expected Behavior
When the subcommand writes to both `stdout` and `stderr`, the order in which is written to the two of these is preserved.

When wrapping [this example](https://goplay.space/#-HWl4aV2FYY) in `secrethub run` the output is always `123`.

mackenbach:
Wow nice catch! Quesstion I have is: the terminal may show both `stdout` and `stderr` in one output stream, but does it give guarantees as to the order of output? Because the order of `stdout` and `stderr` themselves is preserved, just not when they're mixed.

In other words, is not preserving the order an issue or does it fall within the expected behavior of the terminal (when you mix both output streams to one)?

PR #267, opened 2020-03-23
Masking rewrite

jpcoenen:
Complete rewrite of the masking functionality, with the following goals:

1. Fix output to `stdout` and `stderr` getting mixed when written approximately at the same time. (#196 )
2. Make code more legible, understandable and robust. The original implementation had some different constraints and requirements than what is has evolved to. This has led to a lot of unnecessary complexity, which can be removed by rethinking the design.
3. Reduce the data-dependency on the timing of the masking functionality. The previous implementation had a strong relation between whether input contains secrets and the time consumed by the masker. With the original assumption that only the secret's owner has control over the input of the masker, this was no issue. However, as the masking functionality has broad applicability, it was decided to reduce this data-dependency as much as possible.


internals/cli/masker/writer.go
```diff
@@ -1,225 +1,171 @@
...
-		errCh:           make(chan error, 1),
-		outputTimeoutCh: make(chan struct{}, 1),
-		incomingBytesCh: make(chan []byte, 256),
-		outputCh:        make(chan []maskByte),
+func (m *Masker) AddStream(w io.Writer) io.Writer {
+	s := stream{
+		dest:        w,
+		masker:      m,
+		matches:     map[int64]struct{}{},
+		maxLookback: 1024,
```
SimonBarendse: No need to look back further than the length of the longest string to mask, right? We could compute that once when constructing the masker and use that as the maximum lookback. In most cases, I imagine that being a lot smaller than 1024.

internals/secrethub/run.go
```diff
@@ -99,15 +102,17 @@ func (cmd *RunCommand) Run() error {
...
+			sequences = append(sequences, regexp.MustCompile(regexp.QuoteMeta(val)))
 		}
 	}
 
-	maskedStdout := masker.NewMaskedWriter(cmd.io.Stdout(), valuesToMask, maskString, cmd.maskingTimeout)
-	maskedStderr := masker.NewMaskedWriter(os.Stderr, valuesToMask, maskString, cmd.maskingTimeout)
+	m := masker.Masker{
+		BufferDelay:    time.Millisecond * 100,
+		MatchSequences: sequences,
+	}
```
SimonBarendse: Shall we create these regex sequences in a Masker constructor? No need to expose that we're using regexes internally to match the sequences.

SimonBarendse: That makes it easier to use and for us to change it later (e.g. to improve performance).

internals/cli/masker/writer_test.go
```diff
@@ -1,322 +0,0 @@
-package masker
```
SimonBarendse: There was a lot of value in these tests. Let's make sure we don't lose that.

internals/cli/masker/writer.go
```diff
@@ -1,225 +1,171 @@
...
-					masking = false
-				}
-			}
-			mw.wg.Add(-len(output))
-		case <-time.After(mw.timeout):
-			// send to the timeout channel if not already done so.
-			select {
-			case mw.outputTimeoutCh <- struct{}{}:
-			default:
+		case <-trigger.timer.C:
```
SimonBarendse: Maybe we could check whether the next trigger is also done and combine them? That way, we don't check for matches multiple times when not necessary.

internals/cli/masker/writer.go
jpcoenen: You're right, but I'd rather get rid of the `maxLookback` altogether.

internals/cli/masker/writer_test.go
jpcoenen: Of course. That's an artifact of this being WIP.

jpcoenen:
Should have said this in the PR description: but for now I'm mainly looking for a high-level review. So: is this new structure we want to go ahead with? Then I'll polish it.

internals/cli/masker/writer.go
jpcoenen: This depends on how we want to solve the secret matching logic. If we make that more like the previous implementation (with a stateful matcher), the number of times `flushN()` is called, does not affect the performance.

internals/cli/masker/writer.go
```diff
@@ -1,225 +1,171 @@
...
-	return &MaskedWriter{
-		w:               w,
-		maskString:      maskString,
-		matchers:        matchers,
-		timeout:         timeout,
-		errCh:           make(chan error, 1),
-		outputTimeoutCh: make(chan struct{}, 1),
-		incomingBytesCh: make(chan []byte, 256),
-		outputCh:        make(chan []maskByte),
+func (m *Masker) AddStream(w io.Writer) io.Writer {
```
mackenbach: Had a quick look, one think I noticed is that it looks like the `AddStream` function assumes the source (`io.Reader`) and destination (`io.Writer`) are the same. Consider whether you want to make that assumption or make both configurable.

mackenbach:
Small suggestion: @jpcoenen if you want high level overview, maybe draw a quick diagram of how it works? Even a handdrawn sketch would help.

internals/cli/masker/writer.go
jpcoenen: I don't understand what you mean here :thinking: Afaik there is no `io.Reader` involved here. Which reader are you referring to?

jpcoenen:
Will do :+1: That's coming tomorrow.

internals/cli/masker/writer.go
mackenbach: Ahhh okay, I understood it wrong I think. I thought you read from the thingy you pass to the child process as `stdout` and `stderr` right, but that is actually a custom writer not a reader. The custom writer is passed to the child process and then on every write to the custom writer it writes to the underlying writer (`os.StdOut` or `os.StdErr`) with modifications like masking. 

Consider my comment unsaid. 😊

jpcoenen: What comment?

jpcoenen:
@SimonBarendse @mackenbach I've done quite a rewrite. It might still look very complex. However, I think that it has been reduced to the inherent complexity of the required behaviour. The signature of every function is as simple as I could get it. 

A global overview of how it works (also included a diagram; hope it helps):

## Ingesting frames
1. `stream.Write()` is called whenever the child process writes something to `stdout` or `stderr`. Then this happens: 
    a. `IndexedBuffer.Write()`: we store the incoming bytes (a frame) in a buffer 
    b. `multipleMathcer.Write()`: we directly check for any secrets. 
    b. `stream.addMatch()`: If we find any secrets, we store them in a map with the index of the match (which is incrementing and unqiue over the lifetime of the masker) and the match's length. 
   c. `Masker.registerFrame()`: We also set a timer to flush all the bytes of this frame after `BufferDelay` has passed.

## Flushing frames
`Run()` (executed in a goroutine) continuously waits for timers to expire. if this happens, it calls  `stream.flush()`, which writes the frame to the destination `io.Writer`, while masking any previously found secrets.

internals/cli/masker/matcher.go
```diff
@@ -0,0 +1,91 @@
...
+func (m Matches) Add(index int64, length int) Matches {
+	existing, exists := m[index]
+	if !exists || existing < length {
+		m[index] = length
+	}
+	return m
+}
+
+// multipleMatcher combines multiple sequenceMatchers to check for matches of secrets against any of them.
+type multipleMatcher struct {
```
mackenbach: Does a `multipleMatcher` check for multiple secrets to match against? The name `multiple` is a bit ambiguous. I thought for a second that multiple streams would use the same `multipleMatcher` but that's not the case, right?

internals/cli/masker/stream.go
```diff
@@ -0,0 +1,122 @@
+package masker
+
+import (
+	"bytes"
+	"io"
+	"sync"
+)
+
+// stream is a buffered io.Writer that masks all secrets written on it using a multipleMatcher.
```
mackenbach: Do all streams use the same `multipleMatcher`? Is it relevant to mention the `multipleMatcher` in the docs here?

internals/cli/masker/stream.go
```diff
@@ -0,0 +1,122 @@
...
+	matcher     *multipleMatcher
+	matches     Matches
+	matchesLock sync.Mutex
+}
+
+// Write implements the io.Writer interface for the stream.
+// The written frame is stored in the buffer and it is registered in the Masker to make sure it is flushed from
+// the buffer after the buffer delay has passed.
+// The bytes are also passed to the secret matcher to check for any matches with secrets.
+func (s *stream) Write(p []byte) (int, error) {
```
mackenbach: Are bytes (or frames) always buffered for the same amount of time (the constant buffer delay)? If so, maybe it's good to mention that?

internals/cli/masker/stream.go
```diff
@@ -0,0 +1,122 @@
...
+	n, err := s.buf.Write(p)
+	if n > 0 {
```
mackenbach: Is there a valid case where `n == 0` and you need to `register` it??

internals/cli/masker/stream.go
```diff
@@ -0,0 +1,122 @@
...
+
+// flush n bytes from the buffer and mask any secrets that have been matched.
+func (s *stream) flush(n int) error {
+	startIndex := s.buf.CurrentIndex()
+	endIndex := startIndex + int64(n)
+
+	// Increment the frameIndex before processing matches to avoid adding new matches in the processed frame.
+	for i := startIndex; i < endIndex; i++ {
+		s.matchesLock.Lock()
+		length, exists := s.matches[i]
```
mackenbach: Is this a read only lock? Then it may be beneficial to use a `RWMutex` instead.

SimonBarendse:
Are you still looking for only a high level review?

Based on your diagram, the high level implementation looks good! Looks like the two important jobs, masking and buffering/timeouts are nicely separated.

Is there a need for the bytes to go through the masking first and then in the buffer? Seems like going from the `stdout` write directly to the buffer could simplify it a bit more, because the bytes don't have to pass through the "stream" (masking engine) one more time. But, while this might seem complex in the diagram it might work out pretty smooth in the implementation?

internals/cli/masker/masker.go
```diff
@@ -0,0 +1,103 @@
...
+// have them redacted if any matches are found. Output on all streams is buffered to increase the chance of finding
+// secrets if they are spread across multiple writes.
+//
+// Usage:
+// 1. Create a new Masker using New()
+// 2. Add one more streams using AddStream()
+// 3. Run the Run() method in a separate goroutine
+// 4. After everything has been written to the io.Writers, flush all buffers using Flush()
+type Masker struct {
+	BufferDelay time.Duration
```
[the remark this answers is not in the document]
jpcoenen: Yes. Allows the user to tweak it if desired. The default is a sane one.

jpcoenen:
> Are you still looking for only a high level review?

Always welcome, but I am pretty convinced now that this architecture is more solid than the previous one.

> Is there a need for the bytes to go through the masking first and then in the buffer? Seems like going from the stdout write directly to the buffer could simplify it a bit more, because the bytes don't have to pass through the "stream" (masking engine) one more time. But, while this might seem complex in the diagram it might work out pretty smooth in the implementation?

I don't really get what you mean here. If you think it's important, let's discuss it face-to-face.

internals/cli/masker/stream.go
jpcoenen: It is read-only, but the read is only performed from 1 goroutine, so it makes no difference.

internals/cli/masker/stream.go
jpcoenen: Good question, but an empty frame does nothing. It results in a flush that has a loop that immediately terminates. So there is no need to register it.

internals/cli/masker/masker.go
```diff
@@ -0,0 +1,103 @@
...
+	sequences    [][]byte
+	frames       chan frame
+	flushAllChan chan struct{}
+	err          error
+}
+
+// New creates a new Masker that scans all streams for the given sequences and masks them.
+func New(sequences [][]byte) *Masker {
+	return &Masker{
+		BufferDelay:  time.Millisecond * 100,
```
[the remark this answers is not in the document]
jpcoenen: It's configurable by editing after calling `New()`. I thought that a while back we agreed on that defining single-use constants does not add much value. That's why I did it this way.

internals/cli/masker/masker.go
```diff
@@ -0,0 +1,103 @@
...
+		dest:          w,
+		registerFrame: m.registerFrame,
+		matches:       Matches{},
+		matcher:       newMultipleMatcher(m.sequences),
+	}
+	return &s
+}
+
+// Run continuously flushes the input buffer for each frame for which the buffer delay has passed.
+// This method should be run in a separate goroutine.
```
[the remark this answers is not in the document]
jpcoenen: Rephrased to `This method blocks until Flush() is called.`.

internals/cli/masker/masker.go
```diff
@@ -0,0 +1,103 @@
...
+// If a struct is passed on flushAllChan, all pending frames are flushed to the output and the method returns.
+func (m *Masker) Run() {
```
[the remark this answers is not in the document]
jpcoenen: Good suggestion.

> Also, do 'streams' still accept write calls after Flush has been called? And is that desired (and documented)?
```
// This should be run after all input has been written to the io.Writers of the streams.
```
I'll see whether I can make this a little more explicit.

internals/cli/masker/masker.go
```diff
@@ -0,0 +1,103 @@
...
+			if err != nil {
+				m.handleErr(err)
+			}
+		}
+	}
+}
+
+// Flush all pending frames and wait for this to complete.
+// This should be run after all input has been written to the io.Writers of the streams.
+func (m *Masker) Flush() error {
```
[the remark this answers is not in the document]
jpcoenen: Change to `Stop()`

internals/cli/masker/masker.go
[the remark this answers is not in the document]
jpcoenen: You will get a panic if you write to a masked `io.Writer` after calling `Close()`. I've documented that behaviour. Catching without a panic requires some messy logic because streams would have to know whether the masker has stopped or not.

internals/cli/masker/masker.go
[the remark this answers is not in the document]
mackenbach: Consider making it a parameter to `New` then, as you cannot construct this type yourself and changing it at runtime can create unexpected results.

internals/cli/masker/matcher.go
```diff
@@ -0,0 +1,91 @@
+package masker
+
+// Matches represents a set of sequence matches. The key is the index at which the match is found and the value is the
+// length of the match. The index corresponds to the index of the byte in the BufferedIndex of the stream.
+type Matches map[int64]int
```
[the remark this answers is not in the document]
jpcoenen: If you ever write more than 18.44 exabytes to a stream, you'd have a problem. But that is not gonna happen

internals/cli/masker/matcher.go
jpcoenen: Renamed after discussing on the phone.

internals/cli/masker/stream.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// IndexedBuffer is a goroutine safe buffer that assigns every byte that is written to it with an incrementing index.
+type IndexedBuffer struct {
+	buffer       bytes.Buffer
+	mutex        sync.Mutex
+	currentIndex int64
+}
+
+func (b *IndexedBuffer) CurrentIndex() int64 {
```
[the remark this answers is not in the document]
jpcoenen: Removed it altogether.

internals/cli/masker/stream.go
```diff
@@ -0,0 +1,122 @@
...
+
+func (b *IndexedBuffer) Write(p []byte) (n int, err error) {
+	b.mutex.Lock()
+	defer b.mutex.Unlock()
+	return b.buffer.Write(p)
+}
+
+// UpToIndex pops and returns all bytes in the buffer up to the given index.
+// If all bytes up to this given index have already been returned previously, an empty slice is returned.
+func (b *IndexedBuffer) UpToIndex(index int64) []byte {
```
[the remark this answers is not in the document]
jpcoenen: Made unexported.

internals/cli/masker/stream.go
jpcoenen: Removed the reference to the matcher.

internals/cli/masker/stream.go
jpcoenen: Added `constant`.

internals/cli/masker/matcher.go
```diff
@@ -0,0 +1,91 @@
...
+
+// Add a new match to the map if it does not yet exist or the existing match has a shorter length.
+func (m Matches) Add(index int64, length int) Matches {
```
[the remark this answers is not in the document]
jpcoenen: Made unexported.

internals/cli/masker/matcher.go
```diff
@@ -0,0 +1,91 @@
...
+
+// sequenceMatcher takes in bytes to check whether there is any match with the given sequence.
+type sequenceMatcher struct {
+	sequence     []byte
+	currentIndex int
+}
+
+// WriteByte takes in a new byte to match against.
+// Returns true if the given byte results in a match with sequence
+func (m *sequenceMatcher) WriteByte(in byte) bool {
```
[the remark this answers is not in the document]
jpcoenen: Made unexported.

internals/cli/masker/matcher.go
```diff
@@ -0,0 +1,91 @@
...
+	m.currentIndex -= m.findShift()
+	if m.sequence[m.currentIndex] == in {
+		return m.WriteByte(in)
+	}
+	return false
+}
+
+// findShift checks whether we can also make a partial Match by decreasing the currentIndex .
+// For example, if the sequence is foofoobar, if someone inserts foofoofoobar, we still want to Match.
+// So after the third f is inserted, the currentIndex is decreased by 3 with the following code.
```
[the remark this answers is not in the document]
jpcoenen: Tried to make it simpler, but it is not easy to explain. If you have a suggestion, you're very welcome to share it.

internals/cli/masker/masker.go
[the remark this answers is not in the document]
jpcoenen: Introduced a struct for configuration.

internals/cli/masker/stream.go
```diff
@@ -0,0 +1,123 @@
...
+	referenceTime := time.Now()
+
+	n, err := s.buf.write(p)
+
+	for index, length := range s.matcher.write(p[:n]) {
+		s.addMatch(index, length)
+	}
+
+	if n > 0 {
+		s.registerFrame(s, referenceTime.Sub(time.Now()), n)
```
golangcibot: S1024: should use time.Until instead of t.Sub(time.Now()) (from `gosimple`)

internals/cli/masker/stream.go
```diff
@@ -0,0 +1,122 @@
...
+
+	for index, length := range s.matcher.Write(p[:n]) {
+		s.addMatch(index, length)
+	}
+
+	return n, err
+}
+
+// addMatch adds the match of a secret at the given index and with the given length to the map of matches if the
+// associated bytes have not yet been written to the destination.
```
mackenbach: Consider adding a tiny comment here why it is important that the bytes have not been written to the destination yet. We talked about this and there is actual important insight behind this I think is good to document for future readers of the code.

internals/cli/masker/matcher.go
```diff
@@ -0,0 +1,90 @@
...
+
+// sequenceDetector detects if a sequence is present in the bytes it receives.
+type sequenceDetector struct {
+	sequence []byte
+	index    int
+}
+
+// writeByte takes in a new byte to match against.
+// Returns true if the given byte results in a match with sequence
+func (m *sequenceDetector) writeByte(in byte) bool {
```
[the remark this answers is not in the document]
jpcoenen: Oopsie. You're right.

internals/cli/masker/matcher.go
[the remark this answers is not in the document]
jpcoenen: Counterproposal:
```
// findShift checks whether we can also make a partial match by shifting the detector's index, returning the number
// of positions the index should be shifted. If no partial match can be made by shifting, the current index is returned.
// For example, if the sequence to match is "foobar" and "foofoobar" is encountered in the stream we still want to
// trigger a positive match. So after the second "f" character is encountered, findShift returns 3 to indicate the
// index should be decreased by 3.
```

jpcoenen:
>  Does it work nicely with (non-TTY) cases where bytes are not buffered before being written to our writer? If I understand your code correctly, that would mean every frame is one byte long?

I haven't been able to reproduce a situation were bytes actually come in one-by-one into the stream. But even if that happens, it goes OK as long as `BufferDelay/FrameBufferLength < write rate`. If the input rate is higher, the stream will block and the child program cannot write to stdout/stderr anymore.

internals/cli/masker/matcher.go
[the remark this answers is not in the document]
mackenbach: Sweet! 👍 Only thing that's left is to evaulate whether returning current index if no shift makes sense. `0` or `-1` would sound more intuitive from a function signature perspective. Then the comparison whether or not it needs to shift doesn't depend on extra state. Small thing though.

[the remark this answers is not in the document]
jpcoenen: I think returning the index makes perfect sense. It just means that you have to shift completely to the beginning. Returning anything different would make the function's behaviour incosistent.

internals/cli/masker/masker.go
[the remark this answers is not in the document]
mackenbach: Should we catch that error? You could wrap the streams with a very simple `io.Writer` interface that does exactly that one check.  Can be a simple `writeFunc` or something.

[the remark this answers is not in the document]
jpcoenen: Hmm, interesting approach. Will look into it.

internals/cli/masker/matcher.go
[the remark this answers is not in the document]
mackenbach: Just read the usage of `findShift`, it indeed seems like this is the intuitive usage and seeing as it's the only caller let's keep it this way.

mackenbach:
> The only problem I could identify so far is one that is also in the current implementation: interactive utilities do not appear to work correctly. For `htop` this can be fixed by not masking `stderr`.

Do we want to fix the prompting issue? Or is documenting that interactive cases are hard to mask and recommend using `--no-masking`? One of the specific cases I can remember is wrapping `terraform apply` in `secrethub run`.

jpcoenen:
>  One of the specific cases I can remember is wrapping terraform apply in secrethub run.

`terraform run` should not be a problem. It's not the kind of interactive I mean. I mean the ones like `htop` or `vim` that redraw the screen.

internals/cli/masker/masker.go
```diff
@@ -0,0 +1,139 @@
...
+	frames      chan frame
+	stopChan    chan struct{}
+	err         error
+}
+
+// Options for configuring masking behavior.
+type Options struct {
+	// DisableBuffer completely disables the buffering of the masker. This increases output responsiveness
+	// but also increases the chance of a secret not being masked.
+	DisableBuffer bool
```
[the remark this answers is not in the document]
jpcoenen: Linked it to a CLI flag.

mackenbach:
@jpcoenen I recall you saying something like the code could be tested better. Is that something you still plan to pick up in this PR?

internals/cli/masker/masker.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// BufferDelay is the constant duration for which input to a stream is buffered. A higher value increases
+	// the chance of secrets being detected for masking. Especially when writes have a variable delay between them,
+	// for example in the case data arrives over an unstable network connection.
+	// Defaults to 50ms if not set.
+	BufferDelay time.Duration
+
+	// FrameBufferLength is the number of frames that can be in the buffer simultaneously.
+	// If the frame buffer is full, writing to a stream blocks until there is space.
```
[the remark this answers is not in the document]
jpcoenen: Hmm. Interesting approach. I think there is definitely something to say for this. But it also introduces a problem: if you do this, there is a direct path from `Write()` to `flush()`, from which the time of the matching logic could theoretically be deduced.

internals/cli/masker/stream.go
```diff
@@ -0,0 +1,124 @@
...
+	matchesLock sync.Mutex
+}
+
+// Write implements the io.Writer interface for the stream.
+// The written frame is stored in the buffer and it is registered in the Masker to make sure it is flushed from
+// the buffer after the constant buffer delay has passed.
+// The bytes are also passed to the secret matcher to check for any matches with secrets.
+func (s *stream) Write(p []byte) (int, error) {
+	// Save the current time to compensate for the time taken to match for secrets.
+	referenceTime := time.Now()
```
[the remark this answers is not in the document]
jpcoenen: > What happens if the buffer delay is smaller than the time taken to match for secrets? 

A timer in the past triggers immediately. 

> Does matching for secrets take longer if something is a secret? 

The matching logic itself currently is not time-insensitive. 

>  Can you use this as a timing oracle? Should we therefore enforce a minimum buffer delay?

Hmm. This indeed is a problem. A minimum delay won't help here. You could just increase the size of a write until you go over the buffer period. Did a quick test and for my laptop, I get near 50ms for the matching logic for 100 secrets and a 50KB write. So hitting this is possible.

A possible solution would be to chop up the write in smaller chunks, so the matching logic takes way less than 50ms for every chunk.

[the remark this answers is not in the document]
jpcoenen: Scrap that solution. That re-introduces the problem of the output getting scrambled.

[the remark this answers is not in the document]
mackenbach: Does setting a practical size limit for secrets that need to be masked (e.g. max 1KB) solve some of your problems?

internals/secrethub/run.go
```diff
@@ -81,7 +80,8 @@ func (cmd *RunCommand) Register(r command.Registerer) {
 	clause.Alias("exec")
 	clause.Arg("command", "The command to execute").Required().StringsVar(&cmd.command)
 	clause.Flag("no-masking", "Disable masking of secrets on stdout and stderr").BoolVar(&cmd.noMasking)
-	clause.Flag("masking-timeout", "The maximum time output is buffered. Warning: lowering this value increases the chance of secrets not being masked.").Default("1s").DurationVar(&cmd.maskingTimeout)
+	clause.Flag("no-output-buffering", "Disable output buffering. This increases output responsiveness, but decreases the probability that secrets get masked.").BoolVar(&cmd.maskerOptions.DisableBuffer)
+	clause.Flag("masking-buffer-period", "The time period for which output is buffered. A higher value increases the probability that secrets get masked but decreases output responsiveness.").Default("50ms").DurationVar(&cmd.maskerOptions.BufferDelay)
```
SimonBarendse: What's the difference between `--no-output-buffering` and `--masking-buffer-period 0`?

internals/cli/masker/matcher.go
```diff
@@ -0,0 +1,135 @@
...
+	// 	if d.sequence[d.index] == in {
+	// 		d.index++
+	//
+	//		if d.index == len(d.sequence) {
+	//			d.index = 0
+	//			return true
+	//		}
+	//		return false
+	//	} else if d.sequence[0] == in{
+	//		d.index = d.offset
```
SimonBarendse: What's `d.offset` about? I'd expect the index to go to `1` (0 + 1)

SimonBarendse: Also, I see `d.offset+1` used in the actual code, is that correct?

internals/cli/masker/matcher.go
```diff
@@ -0,0 +1,135 @@
...
+	//	} else {
+	//      d.index = 0
```
SimonBarendse: Why does it start at zero and not at `d.offset` here?

internals/cli/masker/matcher_test.go
```diff
@@ -14,6 +18,8 @@ func TestMultipleMatcher(t *testing.T) {
 		[]byte("test"),
 		[]byte("test but longer"),
 		[]byte("another first"),
+		[]byte("112"),
+		[]byte("22222222223"),
```
mackenbach: how does it work for repeating repetitions, e.g. `111211121112`?

internals/secrethub/run.go
jpcoenen: `--no-output-buffering` removes the buffer -> input blocks until flushed til output.
`--masking-buffer-period 0` directly flushes output, but there still is a buffer.

internals/cli/masker/matcher.go
jpcoenen: Well.. It's a little difficult to explain, but I'll try my best to explain it.

Take the case where we want to detect the sequence `ab ab ab c`. In that case, we'd end up with the following detectors in the matchers (where `^` points to the offset):
```
  ab ab ab c
 ^
  ab ab ab ab c
    ^
  ab ab ab ab ab c
       ^
```
The first one is responsible for detecting  `ab ab ab c` and `ab ab ab ab ab ab c` (and anything else that has a multiple of 3x`ab ` in front of it).

The second one is responsible for detecting  `ab ab ab ab c` and `ab ab ab ab ab ab ab c` (and anything else that has `ab ` plus a multiple of 3x`ab ` in front of it).

And so one for the third detector.

--- 

Now let's take the scenario where the input is `ab ab ab ab ab ab ab c`. This is handled by the second detector with the sequence `ab ab ab ab c` and an offset of `1`.

After `ab ab ab ab` is processed, the index is at 12. Then an `a` is checked and this does not correspond to the expected `c`. However, it does correspond to the first character of the sequence. 

If we now reset index to `0`, the remaining input (`ab ab ab c`) will not trigger a detection. However, if we reset the index to offset instead, the remaining input will lead to a match.

--- 

Does that explain it?

internals/cli/masker/matcher.go
jpcoenen: See example above.

internals/cli/masker/matcher_test.go
jpcoenen: Added a case.

internals/cli/masker/matcher.go
SimonBarendse: No, I don't think I'm completely there yet.

I'd expect the offset to be 3 (the length of the part we're skipping over: `ab `).

What does the `offset` indicate, is that a number of repetitions, or a relative index (number of bytes)?

internals/cli/masker/masker.go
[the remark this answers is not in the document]
SimonBarendse: Now that the time of matching is constant, can we change this?

internals/cli/masker/matcher.go
jpcoenen: > I'd expect the offset to be 3 (the length of the part we're skipping over: ab ).

You're right. That was a brainfart on my side.

internals/secrethub/run.go
SimonBarendse: Shall we use `input blocks` or something similar in the `--no-output-buffering` help text to clarify this difference to the user?

florisvdg:
@SimonBarendse should I focus the review on the code or more on the functionality?

mackenbach:
@jpcoenen one question that just hit me: what happens if you have a secret to match that is max size (~512KB if I recall correctly)? Would that create a lot of detectors? Or does it have to do with the starting sequence of the secret?

jpcoenen:
> @jpcoenen one question that just hit me: what happens if you have a secret to match that is max size (~512KB if I recall correctly)? Would that create a lot of detectors? Or does it have to do with the starting sequence of the secret?

It only checks for repetitions of the start of the secret. So the size should not matter, only the degree of repetition. If it starts with 
```
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
That does indeed create >300 detectors. If it starts with:
```
caaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
It does with 1 detector.

Would the first scenario be something we want to guard against?

mackenbach:
Great explanation, thanks! 

> Would the first scenario be something we want to guard against?

No I think it's fine. It would be a very weird edge case and caused by weird behavior on the part of the user. Although we could implement a limit, e.g. max 100 repeating first characters, and document that if the limit is exceeded it isn't guaranteed to mask all (partial) matches?

It's a tiny edge case though. We could also implement it when someone runs into an issue. Right now, the way I get it is that it still masks fine, just the memory (and CPU?) footprint increase hugely, right? That's acceptable for a very weird edge case for now I'd say.

jpcoenen:
> It's a tiny edge case though. We could also implement it when someone runs into an issue. Right now, the way I get it is that it still masks fine, just the memory (and CPU?) footprint increase hugely, right? That's acceptable for a very weird edge case for now I'd say.

Yep. Probably becomes unusable. I agree with your conclusion.

jpcoenen:
Updated the diagram:

jpcoenen:
I have created 2 issues for problems with interactive applications, one on Windows and one on Linux:
https://app.clickup.com/t/4ekmc2
https://app.clickup.com/t/4ekjn8

These problems already existed and seem to be only a little/not related to masking. So I suggest to pick them up later, since these kinds of tools are not the main use-case of `secrethub run`.

mackenbach:
Solid, with the new diagram you really see how a very complex technical problem can be solved elegantly with no more than necessary complexity. Very good stuff guys.

internals/cli/masker/stream.go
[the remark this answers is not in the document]
jpcoenen: This should be fixed now by reducing the data-dependency on the timing of the `writeByte()` function.

internals/cli/masker/masker.go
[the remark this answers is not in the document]
jpcoenen: I am not sure... because it can create a little weird flow/dependency in the code. 

My guess would be to implement this in the following way. Replace the channel read in `Start()` with this:
```golang
select {
	case <-m.forceFlush:
	case <-trigger.timer.C:
}
```
And replace `registerFrame()` with the following:
```
	frame := frame{
		length: l,
		stream: s,
		timer:  time.NewTimer(offset + m.bufferDelay),
	}
	select {
	case m.frames <- frame:
	default:
		m.forceFlush <- struct{}{}
		m.frames <- frame
	}
```

---

In most cases, this would lead to the timer being skipped when the `frames` channel is full. However, if `Start()` is not waiting for a timer but blocked in `trigger.stream.flush()` (for example because the stream's destination is blocked), this means a message is added to `m.forceFlush` that is not picked up directly. 

In the next iteration of `Start()` a message is popped from `m.frames`, freeing up a place in the channel and after that, the timer is also directly skipped. Skipping the timer is not necessary anymore, causing the wrong timer to be skipped. If you have >1024 streams in a Masker I think this could even lead to a deadlock (not that this is a realistic scenario, but it signals that the implementation is not correct).

---

Another solution would be to make sending a message on `m.forceFlush` non-blocking:
```golang
select {
	case m.skipFrameTimer <- struct{}{}:
	default:
}
m.frames <- frame
```

This goes wrong in another scenario: if two frames are registered at (roughly) the same time and `Start()` is waiting for a timer to expire. In this case, a timer is only canceled once, though for both frames the timer should be canceled.

--- 

My suggestion would be not to implement this right now, as I think the practical problems of not implementing it are relatively small. At the same time I have the feeling that trying to implement this, leads to some incorrect code that can cause bugs in the future.

My explanation might not be 100% clear. Mainly wrote it down as a reference to myself. . If anything needs further explanation, let's discuss it in a video call.

internals/secrethub/run.go
jpcoenen: Do you have a suggestion on how to phrase this that it makes things clearer? Just saying that the "input blocks" is confusing when you're reading the flag description ("what input? What does it have to do with _output_ buffering?"). 

I have the feeling that it quickly makes things confusing and that people who care about this, can derive this information from the first part ("disable buffering").

internals/cli/masker/masker.go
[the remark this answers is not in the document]
mackenbach: Jep good call. We can always add more sophistication later. It's already pretty polished I feel and the code is elegant enough to allow for these types of changes if and when we need them.

internals/cli/masker/matcher.go
mackenbach: If it's complex, consider two things:

- Have a quick call to explain.
- Improve the docs accordingly.

mackenbach:
Okay stupid question that might fuck up your day: how does the `masker.Masker` depend on the encoding of the `sequences [][]byte` passed to the constructor AND the encoding of the streams (`stdout` and `stderr`)?

For instance, we've encountered a few times UTF16 encoding on Windows. Now I don't know what encoding windows uses for `stdout` and `stderr`, but I think it's worth checking for a second:

1. Is there a real scenario where encoding is different than we're used to?
2. What would be the impact if that occurred?
3. Do we want to guard against that, either with code or with documentation?

internals/cli/masker/masker.go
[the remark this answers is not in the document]
jpcoenen: Getting this solid is not as easy as I thought. For example, what happens if your in a write when you stop the masker? At the moment the behaviour is undefined because you're using the masker wrong. If we start guarding for this, I feel like we should do it properly, but that again adds some complexity. Is that what we want?

jpcoenen:
> Okay stupid question that might fuck up your day: how does the `masker.Masker` depend on the encoding of the `sequences [][]byte` passed to the constructor AND the encoding of the streams (`stdout` and `stderr`)?
> 
> For instance, we've encountered a few times UTF16 encoding on Windows. Now I don't know what encoding windows uses for `stdout` and `stderr`, but I think it's worth checking for a second:
> 
>     1. Is there a real scenario where encoding is different than we're used to?
> 
>     2. What would be the impact if that occurred?
> 
>     3. Do we want to guard against that, either with code or with documentation?

That's a good point. At the moment, if something is written to `stdout` in `UTF-16`, it is not masked. We could cover this by introducing a `--masking-encoding` flag. But I feel that is overkill for now. You are right that we encountered problems with this on Windows, but I don't think that's relevant in this case. The program itself determines what encoding is used in the output. All programs/languages I tested use `ASCII` or `UTF-8`, which both seem to work fine.

Unicode characters themselves are supported. `ⓗⓔⓛⓛⓞ ⓣⓗⓔⓡⓔ ` gets masked normally.

mackenbach:
Awesome, good balanced response. 

> Unicode characters themselves are supported. ⓗⓔⓛⓛⓞ ⓣⓗⓔⓡⓔ gets masked normally.
Do we want to add a test for good measure?

> For now, we could ask to the docs that only UTF-8 output gets masked.
Yes let's do that.

internals/cli/masker/masker.go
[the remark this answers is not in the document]
mackenbach: Yes, there's definitely a tradeoff here. Consider adding to the docs that writing to any of the writers (/streams?) after calling `Masker.Close()` causes a panic. I think that should cover it well enough for now and we can add complexity when we find a valid use case for it.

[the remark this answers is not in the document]
jpcoenen: That's already there https://github.com/secrethub/secrethub-cli/pull/267/files#diff-331cd1ff2b2231141462b60c73817d48R108

internals/secrethub/run.go
SimonBarendse: Yes, I think you're right. It's very hard to get this across in simple words.

Let's leave it out indeed. Wouldn't want to confuse the masses to explain it for the few people that actually need this behavior.

internals/cli/masker/matcher.go
SimonBarendse: Ah okay, than I actually got it :smile: Perfect

SimonBarendse:
Before shipping, could you still address this @jpcoenen ?

> >  Unicode characters themselves are supported. ⓗⓔⓛⓛⓞ ⓣⓗⓔⓡⓔ gets masked normally.
>
> Do we want to add a test for good measure?

mackenbach:
PR descriptions get committed, right? If so, please edit it so it's a bit more helpful for later.

jpcoenen:
@SimonBarendse this means you tested it?

SimonBarendse:
> @SimonBarendse this means you tested it?

Yes

PR #269, opened 2020-03-24
Audit log bash pagination

Marton6:
Redirected the output of `secrethub audit` to the terminal pager configured in the OS environment.

The output of the command is formatted in a table that spans the width of the console window and has columns of equal width. Text that does not fit in its corresponding table cell is split into multiple lines.

If no pager is configured and neither `less` nor `more` is available the command prints 100 lines from the audit log followed by a descriptive error.

The `--json` flag makes the command output data in json format (through the paginator).


internals/secrethub/audit.go
```diff
@@ -150,6 +197,97 @@ func (cmd *AuditCommand) iterAndAuditTable() (secrethub.AuditEventIterator, audi
...
+	cmd.Stdout = outputWriter
+	cmd.Stderr = os.Stderr
+
+	err = cmd.Start()
+	if err != nil {
+		return nil, err
+	}
+	done := make(chan struct{}, 1)
+	go func() {
+		cmd.Wait()
```
golangcibot: Error return value of `cmd.Wait` is not checked (from `errcheck`)

internals/secrethub/audit.go
```diff
@@ -1,41 +1,68 @@
...
 }
 
 // Register registers the command, arguments and flags on the provided Registerer.
 func (cmd *AuditCommand) Register(r command.Registerer) {
-	clause := r.Command("audit", "Show the audit log.")
+	clause := r.Command("audit", "Show the audit log.\nIf the output of the command is parsed by a script the --json flag must be used.")
 	clause.Arg("repo-path or secret-path", "Path to the repository or the secret to audit "+repoPathPlaceHolder+" or "+secretPathPlaceHolder).SetValue(&cmd.path)
-	clause.Flag("per-page", "number of audit events shown per page").Default("20").IntVar(&cmd.perPage)
+	clause.Flag("per-page", "number of audit events shown per page").Default("20").Hidden().IntVar(&cmd.perPage)
+	clause.Flag("json", "output the audit log in json format").BoolVar(&cmd.json)
```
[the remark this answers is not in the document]
mackenbach: Finally, start the description with a capital letter please.

internals/secrethub/audit.go
```diff
@@ -63,13 +90,34 @@ func (cmd *AuditCommand) run() error {
...
+		}
+		formatter = newColumnFormatter(terminalWidth, auditTable.columns())
+	}
+
+	paginatedWriter, err := cmd.newPaginatedWriter(os.Stdout)
+	if err == errPagerNotFound {
+		paginatedWriter = newFallbackPager(os.Stdout)
+	} else if err != nil {
+		return err
+	}
```
mackenbach: This looks a bit weird to me. Why would the `newPaginatedWriter` not also include this logic for falling back to a `fallbackPager`?

Also, consider your naming: what is the difference between a `PaginatedWriter` and a `Pager`? They seem to do the same, so consider naming them consistently.

internals/secrethub/audit.go
```diff
@@ -63,13 +90,34 @@ func (cmd *AuditCommand) run() error {
...
+	defer paginatedWriter.Close()
+
+	if formatter.printHeader() {
```
mackenbach: The function name `printHeader` suggests that it does something, but it actually only checks if it should print the header, right? So consider naming it `shouldPrintHeader` instead.

internals/secrethub/audit.go
```diff
@@ -63,13 +90,34 @@ func (cmd *AuditCommand) run() error {
...
+		header, err := formatter.formatRow(auditTable.header())
+		if err != nil {
+			return err
+		}
+		fmt.Fprintln(paginatedWriter, header)
+	}
```
mackenbach: This printing logic is exactly the kind of stuff I'd expect in a tabled formatter or a tabled writer or something like that. It's a bit weird that we're abstracting all the formatting and printing logic away to formatters and pagers, but then end up putting custom logic in this function.

internals/secrethub/audit.go
```diff
@@ -1,41 +1,68 @@
...
 	"github.com/secrethub/secrethub-cli/internals/cli/ui"
 	"github.com/secrethub/secrethub-cli/internals/secrethub/command"
 	"github.com/secrethub/secrethub-go/pkg/secrethub"
-	"github.com/secrethub/secrethub-go/pkg/secrethub/iterator"
 
 	"github.com/secrethub/secrethub-go/internals/api"
 )
 
+var (
+	errPagerNotFound = errors.New("no terminal pager available. Please configure a terminal pager by setting the $PAGER environment variable or install less or more")
```
mackenbach: Consider putting quotes around `less` and `more`

mackenbach:
Also did a quick user test and it works like a charm! Really love it. You just can't beat an OS native experience like `less` or `more`. 

One weird thing I ran into is that I could hit the right arrow unlimited amount of times to scroll horizontally to infinity. Not sure if I'm just a `less` noob or this is a bug.

mackenbach:
Also, I forgot to write: @Marton6 impressive chunk of functionality you've added! Really one of the bigger PRs you've made, both in code as in impact. 

Would be interesting to think where else we'd want to use this paging behavior.

internals/secrethub/audit.go
```diff
@@ -63,13 +90,34 @@ func (cmd *AuditCommand) run() error {
...
 	}
 
-	tabWriter := tabwriter.NewWriter(cmd.io.Stdout(), 0, 4, 4, ' ', 0)
-	header := strings.Join(auditTable.header(), "\t") + "\n"
-	fmt.Fprint(tabWriter, header)
+	var formatter tableFormatter
+	if cmd.json {
+		formatter = newJSONFormatter(auditTable.header())
+	} else {
+		terminalWidth, err := cmd.terminalWidth(int(os.Stdout.Fd()))
```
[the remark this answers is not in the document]
Marton6: Weirdly, I encountered strange behavior when using the `Stdout` from `ui.io`.

internals/secrethub/audit.go
```diff
@@ -1,41 +1,68 @@
...
+			w, _, err := terminal.GetSize(fd)
+			return w, err
+		},
 	}
 }
 
 // Register registers the command, arguments and flags on the provided Registerer.
 func (cmd *AuditCommand) Register(r command.Registerer) {
-	clause := r.Command("audit", "Show the audit log.")
+	clause := r.Command("audit", "Show the audit log.\nIf the output of the command is parsed by a script the --json flag must be used.")
```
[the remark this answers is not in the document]
Marton6: - I will add a new line
- This was done with the intent of signaling the user that the tabular output is not meant for scripting as it will not be kept backwards compatible. I'm not sure if putting it in the `--json` flag description will achieve the same result, so I'll rephrase this sentence.

internals/secrethub/audit.go
```diff
@@ -150,24 +329,180 @@ func (cmd *AuditCommand) iterAndAuditTable() (secrethub.AuditEventIterator, audi
 	return nil, nil, ErrNoValidRepoOrSecretPath
 }
 
+type pager interface {
+	io.WriteCloser
+	IsClosed() bool
+}
+
+// newPaginatedWriter runs the terminal pager configured in the OS environment
+// and returns a writer to its standard input.
```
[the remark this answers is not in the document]
Marton6: I meant a writer to the standard input of the terminal pager command. Maybe mentioning explicitly that the terminal pager is a command and that the writer is piped to it would make it more intuitive. I didn't mention that since I tried to avoid mentioning how the writer works and only explain what it does.

internals/secrethub/audit.go
```diff
@@ -150,24 +329,180 @@ func (cmd *AuditCommand) iterAndAuditTable() (secrethub.AuditEventIterator, audi
...
+}
+
+// IsClosed checks if the terminal pager process has been stopped.
+func (p *paginatedWriter) IsClosed() bool {
+	if p.closed {
+		return true
+	}
+	select {
+	case <-p.done:
+		p.closed = true
```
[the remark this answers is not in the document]
Marton6: Calling `p.Close()` is not the solution here. The idea is that the user of the paginator can call `IsClosed()` to check if it can stop fetching the audit log entries, and call `Close()` if the program is supposed to end (there are no more entries or there is an error).

The `closed` field is used since we can only read from the `done` channel one time but the user should be able to call `IsClosed()` any number of times and receive correct return values. So once something is read from the channel, we know the pager is closed so we set the `closed` field and never check the channel again. This is the best solution I found for this issue and I think that modifying the `closed` field in `IsClosed()` is not a problem since, from an outside perspective, it does not alter the behavior of the `paginatedWriter` in any (subsequent) calls.

internals/secrethub/audit.go
```diff
@@ -82,30 +130,161 @@ func (cmd *AuditCommand) run() error {
...
+		}
+		_, err = fmt.Fprintln(paginatedWriter, formattedRow)
+		if err != nil {
+			return err
 		}
 	}
+	return nil
+}
+
+type tableFormatter interface {
```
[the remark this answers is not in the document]
Marton6: The idea behind 'passing strings around' was that this way the steps the data goes through are more clear `formatting -> writing`. But indeed making the `formatters` directly write to a `writer` would simplify and decouple the abstraction.

internals/secrethub/audit.go
Marton6: I agree, I think the best solution would be to make `formatters` write the output themeselves.

internals/secrethub/audit.go
Marton6: I wrote this considering that there might be different ways of handling this error so, "whoever" asks for a pager and does not get one should deal with it however they want.

internals/secrethub/audit.go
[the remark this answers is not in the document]
Marton6: > How forward compatible is the `--json` flag? I can imagine us creating a `--format` (or `--output-format` or `--output` or `--output-mode`) flag that we'll want to use with other flags as well. A couple of valid formats can be `text` (or `default`), `table`, `json`, `quiet` (used now to print only ids and I believe to also avoid prompting).
> 
> See also how the AWS CLI does this: https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-output.html

I looked at the AWS CLI, however I was not aware of the variety of output formats we plan on or consider supporting so the `--output` flag for just one alternative output type did not seem practical. I do understand now the forwards compatibility issue, so I will fix this.

internals/secrethub/audit.go
Marton6: And good point. I will name them consistently.

internals/secrethub/audit.go
```diff
@@ -82,30 +130,161 @@ func (cmd *AuditCommand) run() error {
...
+	}
+
+	for i := range res {
+		if res[i] == 0 {
+			res[i] = widthPerColumn
+		}
+	}
+	f.computedColumnWidths = res
+	return res
 }
```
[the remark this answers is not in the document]
Marton6: I added some comments but I'm not sure to what extent to comment the code. Let me know if it is still not clear enough or if you have any other suggestions.

internals/secrethub/audit.go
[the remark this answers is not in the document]
Marton6: I don't think it is tightly coupled, it's just named poorly :sweat_smile:. It is used for formatting the output to both the `table` and the `json` format. I named it `tableFormatter` since it takes the result of the `auditTable` type. I built it this way to make it more reusable:
The `auditTable` returns a table row as a `[]string`, the `tableFormatter` formats it to either `json` or `table` format depending on which instance of the interface it is. This way the `tableFormatter` can be used to format all kinds of table rows not just audit entries.

internals/secrethub/audit.go
```diff
@@ -82,30 +129,180 @@ func (cmd *AuditCommand) run() error {
...
+
+type columnFormatter struct {
+	tableWidth           int
+	writer               io.Writer
+	computedColumnWidths []int
+	columns              []tableColumn
+	didPrintHeader       bool
+}
+
+func (f *columnFormatter) shouldPrintHeader() bool {
```
golangcibot, line 184: func `(*columnFormatter).shouldPrintHeader` is unused (from `unused`)

internals/secrethub/audit.go
```diff
@@ -82,30 +129,180 @@ func (cmd *AuditCommand) run() error {
...
 		return err
 	}
+	_, err = f.writer.Write(formattedRow)
+	return err
+}
 
-	return nil
+// formatRow formats the given table row to fit the configured width by
+// giving each cell an equal width and wrapping the text in cells that exceed it.
+func (f *columnFormatter) formatRow(row []string) ([]byte, error) {
```
golangcibot, line 214: (*columnFormatter).formatRow - result 1 (error) is always nil (from `unparam`)

internals/secrethub/audit.go
```diff
@@ -150,24 +347,188 @@ func (cmd *AuditCommand) iterAndAuditTable() (secrethub.AuditEventIterator, audi
...
+// Close closes the writer to the terminal pager and waits for the terminal pager to close.
+func (p *paginatedWriter) Close() error {
+	err := p.writer.Close()
+	if err != nil {
+		return err
+	}
+	if !p.closed {
+		err = p.cmd.Process.Signal(syscall.SIGINT)
+		if err != nil {
+			p.cmd.Process.Kill()
```
golangcibot, line 409: Error return value of `p.cmd.Process.Kill` is not checked (from `errcheck`)

internals/secrethub/audit.go
```diff
@@ -150,24 +337,188 @@ func (cmd *AuditCommand) iterAndAuditTable() (secrethub.AuditEventIterator, audi
...
+	}
+	if p.closed {
+		return nil
+	}
+	err = p.cmd.Process.Signal(syscall.SIGINT)
+	if err != nil {
+		err = p.cmd.Process.Kill()
+		if err != nil {
+			return err
+		}
```
Marton6, line 404: This should behave well on Windows. But I'd say we should still try it out.

internals/secrethub/audit_repo_test.go
```diff
@@ -102,9 +113,14 @@ func TestAuditRepoCommand_run(t *testing.T) {
...
 					}, nil
 				},
+				format:  formatTable,
 				perPage: 20,
+				terminalWidth: func(int) (int, error) {
+					return 83, nil
+				},
 			},
 			err: testError,
+			out: "",
```
Marton6, line 123: The table header isn't printed anymore in case there is an error before printing the first entry or there are no entries.

Marton6:
Thanks for the feedback, @mackenbach! I addressed all suggestions regarding naming and commenting and some of the architectural changes you recommended.

internals/secrethub/audit.go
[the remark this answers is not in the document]
Marton6: I'm not sure what to name it though.

internals/secrethub/audit.go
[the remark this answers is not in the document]
mackenbach: Yes that's probably because `ui` gives you a TTY if it's in the terminal. But I think we should refrain from using `os.Stdout` and instead adapt the `ui` package to suit your needs. This you should definitely discuss/call about with @SimonBarendse

internals/secrethub/audit.go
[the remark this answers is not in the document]
mackenbach: Haha yes, it's probably a naming issue. Consider this: `table` is the format, just like `json` is the format. What we're actually doing is processing a `list` of `audit events`. So something like `listFmt`, `listFormatter`, `eventListFormatter` or `auditEventListFormatter` could work?

mackenbach:
Great, I think @jpcoenen can take over review from here. Joris, please also weigh in on the unresolved comments.

jpcoenen:
I'll pick it up. Missed Marc's comment about having done a UX review.

mackenbach:
One last thing I noticed in the AWS CLI, is that they have their own `AWS_PAGER` environment variable that allows you to configure the pager used. It feels like that could actually be a feature, being able to specify a pager program without changing the terminal pager program? Or would that be overkill?

We could have it do something like:

1. `SECRETHUB_PAGER` if set, otherwise
2. `PAGER` if set, otherwise
3. `less` if available, otherwise
4. `more` if available, otherwise
5. print 100 lines (right?)

Something to consider. No biggie.

internals/secrethub/audit.go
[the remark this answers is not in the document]
Marton6: Thanks for the suggestions!

internals/secrethub/audit.go
```diff
@@ -82,30 +129,170 @@ func (cmd *AuditCommand) run() error {
...
+func newColumnFormatter(writer io.Writer, tableWidth int, columns []tableColumn) *columnFormatter {
+	return &columnFormatter{writer: writer, tableWidth: tableWidth, columns: columns}
+}
+
+type columnFormatter struct {
+	tableWidth           int
+	writer               io.Writer
+	computedColumnWidths []int
+	columns              []tableColumn
+	didPrintHeader       bool
```
jpcoenen, line 181: Nitpick: if you want to leave out `did`, `printedHeader` or `headerPrinted` works just as well. (feel free to ignore)

internals/secrethub/audit.go
```diff
@@ -82,30 +129,170 @@ func (cmd *AuditCommand) run() error {
...
+			splitCells[j][i] = strings.Repeat(" ", columnWidth)
+		}
+	}
+
+	// convert the grid to a string
+	strRes := strings.Builder{}
+	for j := 0; j < maxLinesPerCell; j++ {
+		strRes.WriteString(strings.Join(splitCells[j], "  ") + "\n")
+	}
+	return []byte(strRes.String())
```
jpcoenen, line 251: This is a really long function. Splitting it up into multiple functions has some benefits:

1. It forces you to limit the contact surface between different sections.
2. It self-describes the functionality.
3. Allows for easier testing (not always necessary)

In this case, I think you already spotted where the functions should be: at the places where you placed comments.

I haven't checked whether it is actually possible in this case, but give it a try.

internals/secrethub/audit.go
```diff
@@ -1,41 +1,73 @@
...
+			w, _, err := terminal.GetSize(fd)
+			return w, err
+		},
 	}
 }
 
 // Register registers the command, arguments and flags on the provided Registerer.
 func (cmd *AuditCommand) Register(r command.Registerer) {
-	clause := r.Command("audit", "Show the audit log.")
+	clause := r.Command("audit", "Show the audit log.\n\nIf the output of the command is parsed by a script an alternative of the default table format must be used.")
```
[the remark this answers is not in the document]
Marton6: Okay, I'll move it to `--output-format`

internals/secrethub/audit.go
```diff
@@ -63,13 +95,28 @@ func (cmd *AuditCommand) run() error {
 		return err
 	}
 
-	tabWriter := tabwriter.NewWriter(cmd.io.Stdout(), 0, 4, 4, ' ', 0)
-	header := strings.Join(auditTable.header(), "\t") + "\n"
-	fmt.Fprint(tabWriter, header)
+	paginatedWriter, err := cmd.newPaginatedWriter(os.Stdout)
```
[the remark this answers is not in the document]
Marton6: I am aware that it should be `cmd.io.Stdout()` but I'm not entirely sure why that option does not work. I'll look into it

internals/secrethub/audit.go
Marton6: Yeah, that sounds better. I'll use `headerPrinted`.

internals/secrethub/audit.go
```diff
@@ -82,30 +129,173 @@ func (cmd *AuditCommand) run() error {
...
-		if i == cmd.perPage {
-			err = tabWriter.Flush()
-			if err != nil {
-				return err
-			}
-			i = 0
+// newJSONFormatter returns a table formatter that formats the given table rows as json.
+func newJSONFormatter(writer io.Writer, fieldNames []string) *jsonFormatter {
+	return &jsonFormatter{
+		encoder: json.NewEncoder(writer),
```
golangcibot, line 149: File is not `gofmt`-ed with `-s` (from `gofmt`)

internals/secrethub/audit.go
Marton6: It should be quite straightforward. Thanks for explaining the motivation behind the requested change. This really helps me avoid repeating mistakes / bad habits such as this one.

Marton6: However I think splitting up this function might also have a disadvantage:
Most of the smaller functions don't make much sense on their own (e.g. `rowToGrid`, or `gridToString`) and can even be distracting. They are also heavily tied to the solution I came up with and are not fundamental to the problem of formatting a table. I think this would make more sense if the formatter was in a different package and these helper functions were not exported, because seeing them as methods on a formatter makes me wonder why they exist.

That being said I understand the advantage of reducing contact surface, making testing easier and increasing readability, so I'll implement the change. But I'm curious about your opinion on this matter, @jpcoenen

internals/secrethub/audit.go
```diff
@@ -63,13 +95,28 @@ func (cmd *AuditCommand) run() error {
...
+	if err == errPagerNotFound {
+		paginatedWriter = newFallbackPaginatedWriter(os.Stdout)
+	} else if err != nil {
+		return err
+	}
+	defer paginatedWriter.Close()
+
+	var formatter tableFormatter
+	if cmd.format == formatJSON {
+		formatter = newJSONFormatter(paginatedWriter, auditTable.header())
```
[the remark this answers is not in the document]
Marton6: Nice catch! I fixed it.

internals/secrethub/audit.go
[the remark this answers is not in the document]
Marton6: Actually, I'm not entirely sure why `cmd.io.Stdout()` is different from `os.Stdout`.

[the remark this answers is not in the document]
jpcoenen: > Actually, I'm not entirely sure why cmd.io.Stdout() is different from os.Stdout either.

What do you means?

internals/secrethub/audit.go
jpcoenen: I understand your concerns.

> They are also heavily tied to the solution I came up with and are not fundamental to the problem of formatting a table.

That's very true. The question is whether that is a problem. The methods you created are unexported, so in my opinion it's not a problem that they are specific to the underlying implementation. If there were exported, it would be another story and it would indeed be really smelly.

As you already note yourself: Go does not have a very tight mechanism for keeping functionality private to objects except for moving it to another package. I regularly run into situations where I'd like something to be private from other functions within the package. However, that's something that Go is not really built for. The main focus of the exported/unexported mechanism seems to be on hiding internal logic to _external_ users.

> I think this would make more sense if the formatter was in a different package and these helper functions were not exported, because seeing them as methods on a formatter makes me wonder why they exist.

That's not a bad idea. If there's a clear cut you can make, this is definitely an option. However, something to always keep in mind is not to create new packages just because you want a few functions/fields to be unaccessible; that quickly leads to the madness of having _many_ very small packages. Nevertheless, in this case it does appear to be contained and reusable logic, so a new package does seem appropriate.

> Most of the smaller functions don't make much sense on their own (e.g. rowToGrid) and can even be distracting. 

I agree on the `rowToGrid` part. Often that's a sign that:

1. The function does not have a clear, single responsibility, hence it cannot be described in a name. Fix: split or restructure the function.
1. The name does not really describe what is going on. What could work in this case is trying to explain to yourself what the code does. If you find yourself saying "and" a lot while doing this, refer to 1.

In this specific case, I think it's the latter. 

One thing I can imagine helps in naming here, is to rename the variable `row` to `cells` (at least if I correctly understand the purpose of this cell). Then the purpose of the function is easier to describe: "fit the content of cells to the given column widths" (again: correct me if I am wrong here). Then the function could be named something like: `fitCellsToColumns` or maybe even `fitToColumns` (because `cells` is implied in the argument). Does that help?

internals/secrethub/audit.go
```diff
@@ -82,30 +129,198 @@ func (cmd *AuditCommand) run() error {
...
+// and the table width.
+func (f *tableFormatter) columnWidths() []int {
+	if f.computedColumnWidths != nil {
+		return f.computedColumnWidths
+	}
+	res := make([]int, len(f.columns))
+
+	// Distribute the maximum width equally between all columns and repeatedly
+	// check if any of them have a smaller maximum width and can be shrunk.
+	// Stop when no columns can be further adjusted.
```
jpcoenen, line 292: I'm having some trouble understanding what's happening in this function. Maybe it helps if these comments are moved to the place where they actually apply.

internals/secrethub/audit.go
```diff
@@ -82,30 +129,198 @@ func (cmd *AuditCommand) run() error {
...
+	adjusted := true
+	columnsLeft := len(f.columns)
+	widthLeft := f.tableWidth - 2*(len(f.columns)-1)
```
jpcoenen, line 295: Where this value comes from could use a comment.

Marton6: Oh, right. I forgot to comment that. :sweat_smile:

internals/secrethub/audit.go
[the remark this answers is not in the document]
Marton6: I mean that I'm not sure why they behave differently in this context. The former works as expected, the latter basically dumps all the entries on the screen and doesn't stop or allow scrolling.

internals/secrethub/audit.go
Marton6: Thanks, this is great feedback. The trade offs are quite clear now. I'll move the pagination logic to a different package, as suggested.

internals/secrethub/audit.go
```diff
@@ -150,24 +337,188 @@ func (cmd *AuditCommand) iterAndAuditTable() (secrethub.AuditEventIterator, audi
...
+	if err == nil {
+		return pager, nil
+	}
+
+	return "", errPagerNotFound
+}
+
+// newFallbackPaginatedWriter returns a pager that closes after outputting a fixed number of lines without pagination
+// and returns errPagerNotFound on the last (or any subsequent) write.
+func newFallbackPaginatedWriter(w io.WriteCloser) pager {
```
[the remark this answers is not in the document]
Marton6: Good point!

internals/secrethub/audit.go
```diff
@@ -1,41 +1,73 @@
 package secrethub
```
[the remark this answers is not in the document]
Marton6: I moved the pagination logic to another package and the formatting to another file in the `secrethub` package. Is that a reasonable approach?

internals/secrethub/audit.go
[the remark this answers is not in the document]
SimonBarendse: Have you had a look at `io.Prompts` as an alternative to `io.Stdout`?

Is this still relevant? Give me a call if you'd like to discuss.

internals/secrethub/audit.go
```diff
@@ -150,24 +337,188 @@ func (cmd *AuditCommand) iterAndAuditTable() (secrethub.AuditEventIterator, audi
...
+// pagerCommand returns the name of the terminal pager configured in the OS environment ($PAGER).
+// If no pager is configured it falls back to "less" than "more", returning an error if neither are available.
+func pagerCommand() (string, error) {
+	var pager string
+	var err error
+
+	pager, err = exec.LookPath(os.ExpandEnv(pagerEnvvar))
+	if err == nil {
+		return pager, nil
+	}
```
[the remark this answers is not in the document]
Marton6: Nice, thanks for the tip :+1:

internals/secrethub/formatter.go
```diff
@@ -153,40 +153,41 @@ func (f *tableFormatter) columnWidths() []int {
...
-			for i := range res {
-				res[i] += widthLeft / len(res)
+			for i := range fixedWidths {
+				fixedWidths[i] += widthLeft / len(fixedWidths)
 			}
 			break
 		}
+		// Recalculate the width/column for the remaining unadjusted columns.
 		widthPerColumn = widthLeft / columnsLeft
 	}
```
Marton6: I wrote a couple more comments and renamed a slice to make the code a bit more clear. I find it quite challenging to determine how detailed the commenting should be so please suggest any further changes if it is still not clear.

internals/secrethub/audit.go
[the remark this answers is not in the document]
jpcoenen: Been playing a little with it and I think I found the reason for this.

My suspicion is the following: some functionality inside the `cmd.NewPaginatedWriter` needs an `*os.File` for its interactive functionality. It obtains this by casting the provided `io.Writer` to  file:
```
writer.(*os.File)
```

This works if `os.Stdout` is used because it is an `*os.File`. However, `cmd.io.Stdout()` is not an `*os.File` it's an `ui.file`, so casting does not work... You can simulate this by (it panics):
```
io.Writer(cmd.io.Stdout()).(*os.File)
```
This casting problem is probably handled gracefully by reverting to non-interactive mode.

The underlying problem is that the `ui.IO` interface is not pretty. It is a weird abstraction that does some magic by embedding `*os.File` and creating some custom interfaces. That probably requires some work to make this work.

internals/secrethub/audit.go
jpcoenen: Yes, it's a lot more readable now :grin:

internals/secrethub/audit.go
```diff
@@ -63,13 +88,28 @@ func (cmd *AuditCommand) run() error {
...
 
-	tabWriter := tabwriter.NewWriter(cmd.io.Stdout(), 0, 4, 4, ' ', 0)
-	header := strings.Join(auditTable.header(), "\t") + "\n"
-	fmt.Fprint(tabWriter, header)
+	paginatedWriter, err := cmd.newPaginatedWriter(os.Stdout)
+	if err == pager.ErrPagerNotFound {
+		paginatedWriter = pager.NewFallbackPager(os.Stdout)
+	} else if err != nil {
+		return err
+	}
```
jpcoenen, line 96: Is there a reason for doing the fallback here or could we move this logic to the `pager.New` functionality?

internals/secrethub/formatter.go
```diff
@@ -0,0 +1,193 @@
+package secrethub
```
jpcoenen, line 1: Should this file be called `formatter.go` or `list_formatters.go`?

internals/secrethub/formatter.go
```diff
@@ -153,40 +153,41 @@ func (f *tableFormatter) columnWidths() []int {
 	if f.computedColumnWidths != nil {
 		return f.computedColumnWidths
 	}
-	res := make([]int, len(f.columns))
+	fixedWidths := make([]int, len(f.columns))
```
jpcoenen, line 156: :+1: for renaming it to a more verbose name. Specifically in this case however, `fixed` can have a double meaning: "adjusted"/"repaired" or "not changing" (a fixed value). Changing it to something like `adjustedWidths` does away with this possible confusion.

Besides that the new comments really improve the legibility of the code :ok_hand: :

internals/secrethub/audit.go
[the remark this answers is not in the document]
jpcoenen: For this to work, you also have to remove the `var` statements at the beginning of the functions and use `:=` instead of `=`. Only then the variables are tightly scoped.

```golang
if pager, err := exec.LookPath(os.ExpandEnv(pagerEnvvar)); err == nil {
	return pager, nil
}
// This does not work because pager is out of scope here:
fmt.Println(pager)
```

internals/secrethub/audit.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 var (
-	errNoSuchFormat = func(format string) error { return errors.New("invalid format: " + format) }
+	errNoSuchFormat = func(format string) error { return errors.New("invalid format: " + format + " (audit.invalid_format)") }
```
jpcoenen, line 23: Is there a reason not to use `errio` here? I know it's old and ugly, but doing this concatenation manually is going to give us some headaches in the future if we're ever moving away from that package.

internals/secrethub/audit.go
jpcoenen: Did you have a look at my point on how to possibly improve the naming of/within `writeToGrid`?

internals/secrethub/audit.go
Marton6: In other PRs I was told to not use `errio` as we're trying to move away from it. (Admittedly, the context was a bit different.). It would be nice to have this written down somewhere so that we keep the code consistent.

Marton6: And I agree, I'll just use errio.

internals/secrethub/audit.go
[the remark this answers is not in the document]
Marton6: Missed that. Thanks for pointing it out!

internals/secrethub/audit.go
Marton6: Yes, because it seemed a bit too specific. I think that we might want to handle not having a terminal pager differently in other situations. Or do you think we should handle it in `New` for now and if we ever need it we can do a small refactor?

internals/secrethub/formatter.go
Marton6: `list_formatters.go` is indeed a better name for it.

internals/secrethub/audit.go
jpcoenen: Okay, that's a very reasonable thought. Is there maybe a middle-ground between this being configurable and not having to do this error check outside the package? Because now we have to duplicate this check (which is very tightly coupled to the fallback mechanism) if we ever want to re-use the fallback functionality.

Some possibilities I could imagine:
```golang
paginatedWriter, err := pager.NewWithFallback(os.Stdout)
```
Or (which I like less):
```golang
paginatedWriter, err := pager.WithFallback(pager.New(os.Stdout)
```

Or maybe something similar.

Marton6: Good point! I also prefer the former solution.

internals/secrethub/audit.go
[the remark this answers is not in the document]
Marton6: After rewriting the `ui` package a bit based on your observation passing `cmd.io.Stdout()` works as expected.

internals/secrethub/audit.go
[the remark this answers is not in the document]
Marton6: After adapting the `ui` package passing `cmd.io.Stdout()` works as expected.

internals/secrethub/pager/pager.go
```diff
@@ -118,15 +128,15 @@ func pagerCommand() (string, error) {
 
 // newFallbackPaginatedWriter returns a pager that closes after outputting a fixed number of lines without pagination
 // and returns errPagerNotFound on the last (or any subsequent) write.
-func NewFallbackPager(w io.WriteCloser) io.WriteCloser {
+func NewFallbackPager(w io.Writer) io.WriteCloser {
```
jpcoenen, line 137: Does this still have to be exported?

Marton6: No, it doesn't. Made it unexported

Marton6:
1. I made the command use the RFC format for timestamps when outputting json. The only drawback of this is that user can't output json with human readable timestamps anymore.
2. Piping the table output to grep does not work since if the `stdout` of the audit command is piped we can no longer retrieve the width of the terminal and thus can no longer align columns correctly. In such cases we default to a terminal width of 80 which breaks most values in multiple rows. The only solution I can think of is to not align columns and not break values when output is piped.
3. I think we could encourage users to not fetch the entire log by implementing a timestamp based filter that can be configured from the CLI but defaults to only fetching the logs for the last two months (for example). This way users will only fetch the entire log if they really intend to. I imagine in most situations users only care about recent events or events that happened in a specific time interval anyways. To not break backwards compatibility, we can either implement time based filtering in the audit command and change it to the proper, server side audit filtering when that one is implemented or just implement the audit filters on the server before releasing the pagination. 

What do you think?

jpcoenen:
Please note that #274 was reverted with #279. #277 should fix the build problem introduced in #274 and also provides a new signature for the `ui` package. Check it out.

jpcoenen:
> I made the command use the RFC format for timestamps when outputting json. The only drawback of this is that user can't output json with human readable timestamps anymore.

:+1: Don't think that's a big problem.

> Piping the table output to grep does not work since if the stdout of the audit command is piped we can no longer retrieve the width of the terminal and thus can no longer align columns correctly. In such cases we default to a terminal width of 80 which breaks most values in multiple rows. The only solution I can think of is to not align columns and not break values when output is piped.

I think the latter proposal might work a little better. Give it a try and see if it results something that's workable.

> I think we could encourage users to not fetch the entire log by implementing a timestamp based filter that can be configured from the CLI but defaults to only fetching the logs for the last two months (for example). This way users will only fetch the entire log if they really intend to. I imagine in most situations users only care about recent events or events that happened in a specific time interval anyways. To not break backwards compatibility, we can either implement time based filtering in the audit command and change it to the proper, server side audit filtering when that one is implemented or just implement the audit filters on the server before releasing the pagination.

The main issue is in the case when someone is not using the pagination of `less` and starts processing all possible audit log entries. In that case, I think it would be best to have a sensible limit (e.g. 1000 entries) to avoid processing an extremely long list of entries by accident. However, I do not think we want any limit when not piping the output because in that case `less` takes care of this problem. So we could have a limit that kicks in when output is piped. If we do this, it should be well documented and there should be some way to work around this.

Marton6:
Thanks for the feedback and the ideas, @jpcoenen!

The unaligned output works well when the output is piped. 

For limiting the output but allowing the user to work around the limit the most reasonable solution I came up with is adding a `--limit` flag. The only disadvantage of this is that we expose a new flag which needs to be kept backwards compatible.

mackenbach:
Is it a `--limit` or a `--max-results` flag? Not saying it should be one or the other, just to think about the naming and whether it makes sense and aligns with common conventions.

Marton6:
It seems that there is no common convention for this. The most common names for these flags are: `--max-count`, `--max-items`, `--last` and `--max-results`. I'll use `--max-results` as I think it fits best in this context.

PR #274, opened 2020-04-07
Refactor ui package

Marton6:
Refactored the `ui` package to avoid wrapping `os.Stdout` and `os.Stdin` and ease mocking.

### DIFF
--- a/internals/cli/ui/ask.go
+++ b/internals/cli/ui/ask.go
@@ -51,7 +51,7 @@ func AskWithDefault(io IO, question, defaultValue string) (string, error) {
 
 // AskSecret prints out the question and reads back the input,
 // without echoing it back. Useful for passwords and other sensitive inputs.
-func AskSecret(io IO, question string) (string, error) {
+func AskSecret(io IO, passwordReader PasswordReader, question string) (string, error) {
 	promptIn, promptOut, err := io.Prompts()
 	if err != nil {
 		return "", err
@@ -62,14 +62,14 @@ func AskSecret(io IO, question string) (string, error) {
 		return "", err
 	}
 
-	raw, err := promptIn.ReadPassword()
+	raw, err := passwordReader.Read(promptIn)
 	if err != nil {
 		return "", ErrReadInput(err)
 	}
 
 	fmt.Fprintln(promptOut, "")
 
-	return string(raw), nil
+	return raw, nil
 }
 
 // AskMultiline prints out the question and reads back the input until an EOF is reached.
@@ -145,14 +145,14 @@ func ConfirmCaseInsensitive(io IO, question string, expected ...string) (bool, e
 // the answers still haven't matched after trying n times, the error
 // ErrPassphrasesDoNotMatch is returned. For the empty answer ("") no
 // confirmation is asked.
-func AskPassphrase(io IO, question string, repeatPhrase string, n int) (string, error) {
+func AskPassphrase(io IO, passwordReader PasswordReader, question string, repeatPhrase string, n int) (string, error) {
 	_, promptOut, err := io.Prompts()
 	if err != nil {
 		return "", err
 	}
 
 	for i := 0; i < n; i++ {
-		answer, err := AskSecret(io, question)
+		answer, err := AskSecret(io, passwordReader, question)
 		if err != nil {
 			return "", err
 		}
@@ -161,7 +161,7 @@ func AskPassphrase(io IO, question string, repeatPhrase string, n int) (string, 
 			return answer, nil
 		}
 
-		confirmed, err := AskSecret(io, repeatPhrase)
+		confirmed, err := AskSecret(io, passwordReader, repeatPhrase)
 		if err != nil {
 			return "", err
 		}
--- a/internals/cli/ui/ask.go
+++ b/internals/cli/ui/ask.go
@@ -62,7 +62,7 @@ func AskSecret(io IO, passwordReader PasswordReader, question string) (string, e
 		return "", err
 	}
 
-	secret, err := passwordReader.Read(promptIn)
+	secret, err := passwordReader.read(promptIn)
 	if err != nil {
 		return "", ErrReadInput(err)
 	}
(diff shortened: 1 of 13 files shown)


internals/cli/ui/io.go
```diff
@@ -65,50 +67,58 @@ func (o UserIO) Prompts() (Reader, Writer, error) {
...
+	Read(reader io.Reader) (string, error)
 }
 
-// file implements the Reader and Writer interface.
-type file struct {
-	*os.File
+type passwordReader struct{}
+
+// NewPasswordReader returns a reader that reads a string from the terminal without echoing the user input.
+func NewPasswordReader() *passwordReader {
```
golangcibot, line 85: exported func NewPasswordReader returns unexported type *github.com/secrethub/secrethub-cli/internals/cli/ui.passwordReader, which can be annoying to use (from `golint`)

internals/cli/ui/io.go
```diff
@@ -65,50 +67,58 @@ func (o UserIO) Prompts() (Reader, Writer, error) {
...
+
+// NewPasswordReader returns a reader that reads a string from the terminal without echoing the user input.
+func NewPasswordReader() PasswordReader {
+	return &passwordReader{}
 }
 
-// ReadPassword reads from a terminal without echoing back the typed input.
-func (f file) ReadPassword() ([]byte, error) {
+// Read reads one line of input from the terminal without echoing the user input.
+func (pr *passwordReader) Read(r io.Reader) (string, error) {
```
SimonBarendse, line 90: Do we want to use these outside of the`ui` package? Consider making them private. It's functionality is exposed via `AskPassphrase` and `AskSecret`.

internals/cli/ui/ask.go
```diff
@@ -51,7 +51,7 @@ func AskWithDefault(io IO, question, defaultValue string) (string, error) {
 
 // AskSecret prints out the question and reads back the input,
 // without echoing it back. Useful for passwords and other sensitive inputs.
-func AskSecret(io IO, question string) (string, error) {
+func AskSecret(io IO, passwordReader PasswordReader, question string) (string, error) {
```
SimonBarendse, line 54: Do we want the password reader to be configurable? We could also call `NewPasswordReader` internally, keeping usage simple (and backwards compatible).

internals/cli/ui/ask.go
```diff
@@ -145,14 +145,14 @@ func ConfirmCaseInsensitive(io IO, question string, expected ...string) (bool, e
 // the answers still haven't matched after trying n times, the error
 // ErrPassphrasesDoNotMatch is returned. For the empty answer ("") no
 // confirmation is asked.
-func AskPassphrase(io IO, question string, repeatPhrase string, n int) (string, error) {
+func AskPassphrase(io IO, passwordReader PasswordReader, question string, repeatPhrase string, n int) (string, error) {
```
SimonBarendse, line 148: Same thing about making the password reader configurable or not

internals/cli/ui/io.go
```diff
@@ -18,35 +18,37 @@ var (
...
 // IO is an interface to work with input/output.
 type IO interface {
-	Stdin() Reader
-	Stdout() Writer
-	Prompts() (Reader, Writer, error)
+	Stdin() io.Reader
+	Stdout() io.Writer
+	Prompts() (io.Reader, io.Writer, error)
+	IsStdinPiped() bool
+	IsStdoutPiped() bool
```
SimonBarendse, line 25: Do we want to expose these functions via the interface or expose the `isPiped` function instead? (we'd have to change the signature to return files instead of reader/writer though).

internals/cli/ui/ask.go
SimonBarendse: See also the amount of files you have to touch to introduce this change.

internals/cli/ui/io.go
SimonBarendse: Exposing the `isPiped` function helps keeping this interface small and simple.

internals/cli/ui/ask.go
Marton6: I passed the `PasswordReader` to allow injecting it for testing purposes. Without this, one of the `Write` command tests will no longer be possible after this rewrite.

internals/cli/ui/io.go
Marton6: Returning `os.File`s would make mocking difficult, as `os.File` is a complex struct. Therefore I decided it would be simpler and more readable (also in testing code) if we would have this interface.

internals/cli/ui/io.go
Marton6: I agree, I'll make it unexported.

internals/cli/ui/io.go
SimonBarendse: Okay, makes sense. Thanks for the explanation! :smile:

internals/cli/ui/ask.go
SimonBarendse: You mean the `"ask secret success"` and "ask secret error" testcase, right? 

Let's first think about what it is we're trying to test here. It's a unit test for the write command, so we only want to test the logic in the write command, not test the internals of the `AskSecret` function.

If we make the password reader internal to the `AskSecret` function, the test that verifies `AskSecret` handles the output of the password reader correctly should be a test for `AskSecret`. The unit tests for the write command can focus on the interaction between the write command and the `AskSecret` function. Can we mock `AskSecret` instead?

Marton6: Good point! I'll mock `AskSecret` then!

internals/cli/ui/testing.go
```diff
@@ -93,7 +88,8 @@ type FakeWriter struct {
 	Piped bool
 }
 
-// IsPiped returns the mocked Piped.
-func (f *FakeWriter) IsPiped() bool {
-	return f.Piped
+type FakePasswordReader struct{}
+
+func (f FakePasswordReader) read(reader io.Reader) (string, error) {
```
golangcibot, line 93: func `FakePasswordReader.read` is unused (from `unused`)

PR #275, opened 2020-04-07
Move linting to CircleCI

Marton6:


.circleci/config.yml
```diff
@@ -1,5 +1,11 @@
 version: 2
 jobs:
+  lint:
+    docker:
+      - image: circleci/golang:1.13
+    steps:
+      - checkout
+      - run: make lint-ci
```
SimonBarendse, line 8: Have you considered using a docker image that already has the linter installed?

Here's the official golang-ci-lint image: https://hub.docker.com/r/golangci/golangci-lint

You could even consider using an orb, e.g. https://circleci.com/orbs/registry/orb/timakin/golangci-lint (https://github.com/timakin/golangci-lint-orb).

SimonBarendse: I have a slight preference for using a docker image with a `run` step over using an orb, given that the configuration of the `run` step is just as easy as the configuration of the orb (so what's there to gain?) and there's no official orb available.

Marton6: I agree, I'll try using the docker image

PR #276, opened 2020-04-15
Consistent not found errors

Marton6:
In combination with [this PR](https://github.com/secrethub/secrethub-go/pull/183) on the client, this PR makes the CLI always output a `secret not found` error when a secret is not found, as opposed to returning a `resource not found` error when removing and listing secrets and a `secret not found` error otherwise.

### DIFF
--- a/internals/secrethub/list.go
+++ b/internals/secrethub/list.go
@@ -100,9 +100,7 @@ func (cmd *LsCommand) Run() error {
 	secretPath, err := cmd.path.ToSecretPath()
 	if err == nil {
 		versions, err := client.Secrets().Versions().ListWithoutData(secretPath.Value())
-		if api.IsErrNotFound(err) {
-			return ErrResourceNotFound(cmd.path)
-		} else if err != nil {
+		if err != nil {
 			return err
 		}
 
--- a/internals/secrethub/rm.go
+++ b/internals/secrethub/rm.go
@@ -87,8 +87,8 @@ func (cmd *RmCommand) Run() error {
 
 	// Check if the secret exists first so we can return a generic error here instead of ErrSecretNotFound.
 	_, err = client.Secrets().Get(secretPath.Value())
-	if err != nil {
-		return err
+	if api.IsErrNotFound(err) {
+		return ErrResourceNotFound(cmd.path)
 	}
 
 	return rmSecret(client, secretPath, cmd.force, cmd.io)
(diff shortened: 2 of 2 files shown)


internals/secrethub/rm.go
```diff
@@ -87,8 +87,8 @@ func (cmd *RmCommand) Run() error {
 
 	// Check if the secret exists first so we can return a generic error here instead of ErrSecretNotFound.
 	_, err = client.Secrets().Get(secretPath.Value())
-	if api.IsErrNotFound(err) {
-		return ErrResourceNotFound(cmd.path)
+	if err != nil {
+		return err
```
Marton6, line 91: Only not found errors used to be checked and returned here. Was this intentional and can it be changed?

internals/secrethub/list.go
```diff
@@ -100,9 +100,7 @@ func (cmd *LsCommand) Run() error {
 	secretPath, err := cmd.path.ToSecretPath()
 	if err == nil {
 		versions, err := client.Secrets().Versions().ListWithoutData(secretPath.Value())
-		if api.IsErrNotFound(err) {
-			return ErrResourceNotFound(cmd.path)
-		} else if err != nil {
+		if err != nil {
```
Marton6, line 103: This check is now done on the client and an `errSecretNotFound` is returned.

internals/secrethub/rm.go
SimonBarendse: This was intentional and is explained in the comment above:

>  	// Check if the secret exists first so we can return a generic error here instead of ErrSecretNotFound. 

The remove command doesn't know whether the provided path points to a secret. Therefore, if no secret can be found on the given path a more generic error should be returned.

Marton6: Oh, thanks I didn't understand that comment at first. In this case we should still return `ErrResourceNotFound`, right?

SimonBarendse: Yes exactly :+1:

internals/secrethub/list.go
SimonBarendse: Do we want to pass the generic "resource not found" error instead of a "secret not found" error, similar to what was there in the remove command?

SimonBarendse: Because, similar as the `remove` command, the `list` command doesn't know whether the provided path points to a secret. Therefore, if no secret can be found on the given path a more generic error should be returned.

Marton6: Good point. I'll close the PR as no changes are required, then

Marton6:
Closed as the supposed inconsistency is present because the `rm` and `ls` commands can list all types of resources (not just secrets), therefore the returned errors should in fact be `resource not found` errors.

PR #277, opened 2020-04-24
Refactor UI package

mackenbach:
This started out as a fix for the build problems introduced in #274, but ended up as a refactor of the `ui` package. 

The main changes:
- `ui.IO` interface changed from:
```golang
type IO interface {
	Stdin() io.Reader
	Stdout() io.Writer
	Prompts() (io.Reader, io.Writer, error)
	IsStdinPiped() bool
	IsStdoutPiped() bool
}
```
to: 
```golang

type IO interface {
	// Input returns an io,Reader that reads input for the current process.
	// If the process's input is piped, this reads from the pipe otherwise it asks input from the user.
	Input() io.Reader
	// Output returns an io.Writer that writes output for the current process.
	// If the process's output is piped, this writes to the pipe otherwise it prints to the terminal.
	Output() io.Writer
	// Stdin returns the *os.File of the current process's stdin stream.
	Stdin() *os.File
	// Stdin returns the *os.File of the current process's stdout stream.
	Stdout() *os.File
	// Prompts returns an io.Reader and io.Writer that read and write directly to/from the terminal, even if the
	// input or output of the current process is piped.
	// If this is not supported, an error is returned.
	Prompts() (io.Reader, io.Writer, error)
	// ReadSecret reads a line of input from the terminal while hiding the entered characters.
	// Returns an error if secret input is not supported.
	ReadSecret() ([]byte, error)
	// IsInputPiped returns whether the current process's input is piped from another process.
	IsInputPiped() bool
	// IsOutputPiped returns whether the current process's output is piped to another process.
	IsOutputPiped() bool
}
```
This allows users of the interface to explicitly request the underlying `stdin` or `stdout` file when they need it (for example, when spawning a child process). Also the naming of `IsInputPiped` and `IsOutputPiped` better describes the interface than `IsStdinPiped` and `IsStdoutPiped`.

- The interface is actually implemented by different structs (`standardIO`, `windowsIO` and `ttyIO`) instead of a single struct with different fields set. 

This replaces the changes from #274.


jpcoenen:
Build tests are introduced in https://github.com/secrethub/secrethub-cli/pull/278.

jpcoenen:
I think I fixed most of the issues. Could use some manual testing to verify whether it actually works on all platforms.

SimonBarendse:
I believe #274 was also introduced to fix a bug, there is a package (I believe in the standard library) that changes its behavior based on the passed Writer being `os.Stdout`. Before the refactor not `os.Stdout` itself was returned, but a wrapped version. The refactor changed this so that `os.Stdout` would be detected as stdout by the package.

This shows the importance of descriptive and complete PR messages.

@Marton6 Do you remember which bug and package this was about exactly?

internals/cli/ui/io_unix.go
```diff
@@ -2,23 +2,80 @@
...
 
-	return NewStdUserIO()
+	return newStdUserIO()
+}
+
+// Prompts simply returns Stdin and Stdout, when both input and output are
+// not piped. When either input or output is piped, Prompts attempts to
+// bypass stdin and stdout by connecting to /dev/tty on Unix systems when
+// available. On systems where tty is not available and when either input
+// or output is piped, prompting is not possible so an error is returned.
```
mackenbach, line 35: This comment seems outdated and/or may need to be moved (partially) to a different place in the code that is now responsible for this.

internals/cli/ui/io_unix.go
```diff
@@ -28,9 +67,7 @@ func isPiped(file *os.File) bool {
 	}
 
 	return (stat.Mode() & os.ModeCharDevice) == 0
-}
-
-// eofKey returns the key(s) that should be pressed to enter an EOF.
+} // eofKey returns the key(s) that should be pressed to enter an EOF.
```
SimonBarendse, line 78: This seems unintended?

internals/cli/ui/io_windows.go
```diff
@@ -1,24 +1,32 @@
...
+		standardIO: newStdUserIO(),
 	}
+}
 
-	return NewStdUserIO()
+// Stdout returns the standardIO's Output.
+func (o windowsIO) Stdout() io.Writer {
+	if !color.NoColor {
+		return colorable.NewColorable(os.Stdout)
+	}
```
SimonBarendse, line 28: One difference is that this is now evaluated every time `Stdout` is called. Is this a lightweight operation?

internals/cli/ui/io.go
```diff
@@ -18,8 +18,8 @@ var (
 
 // IO is an interface to work with input/output.
 type IO interface {
-	Stdin() io.Reader
-	Stdout() io.Writer
+	Input() io.Reader
+	Output() io.Writer
 	Prompts() (io.Reader, io.Writer, error)
 	IsStdinPiped() bool
 	IsStdoutPiped() bool
```
SimonBarendse, line 27: Now these need to be renamed as well, right?

internals/cli/ui/io.go
```diff
@@ -20,6 +20,8 @@ var (
 type IO interface {
 	Input() io.Reader
 	Output() io.Writer
+	Stdin() *os.File
+	Stdout() *os.File
```
SimonBarendse, line 24: I don't see these being used. (where) are they needed?

internals/cli/ui/io_windows.go
jpcoenen: Good point. Doing this at init now.

internals/cli/ui/io.go
jpcoenen: They are meant for the pagination functionality. But technically `run.go` also needs `Stdout`. So introduced them there.

internals/cli/ui/io.go
jpcoenen: Good point. Done.

internals/cli/ui/io_unix.go
jpcoenen: Yes. Fixed.

internals/cli/ui/io.go
jpcoenen: So this is the main fix. We're exposing `*os.File` through a separate function that can be called when a function needs direct access to `Stdin` or `Stdout`.

SimonBarendse: I'm not sure whether I got this right, but is it about the underlying implementation being checked whether it is `os.Stdout`? If so, couldn't we also return it as an `io.Writer`, as long as the implementation is `os.Stdout` itself and not a wrapped version?

jpcoenen: That would work, yes. But that would not garantuee that it actually is an `*os.File`.

.circleci/config.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - run: make test
   verify-goreleaser:
     docker:
-      - image: goreleaser/goreleaser:v0.127
+      - image: goreleaser/goreleaser:v0.133
```
SimonBarendse, line 25: We should also bump the version in the `Makefile`

SimonBarendse: Ideally, similar to how we did that here: https://github.com/secrethub/terraform-provider-secrethub/pull/82/files#diff-b67911656ef5d18c4ae36cb6741b7965R16

jpcoenen: You mean of golangci-lint? Isn't that compatible with all versions of go `>=1.12`?

SimonBarendse: Oh woops, this is about Goreleaser :see_no_evil: My mistake

You're right, we don't have to update `golangci-lint` :+1:

jpcoenen:
Removed the `bug` label because the PR that introduced this has been reverted. (#279)

Changeset itself is still relevant.

jpcoenen:
Since this is not really useful on itself, I am merging it into #269.

mackenbach:
@jpcoenen feel free to merge when you feel it's ready

.circleci/config.yml
```diff
@@ -1,14 +1,25 @@
...
+      arch:
+        type: string
     docker:
       - image: circleci/golang:1.13
+    steps:
+      - checkout
+      - run: GOOS=<< parameters.os >> GOARCH=<< parameters.arch >> go build ./cmd/secrethub
+  test:
+    docker:
+      - image: circleci/golang:1.14
```
florisvdg, line 22: There's now multiple Go versions, let's pick one.

jpcoenen: That was a result of the merge. Well spotted.

PR #281, opened 2020-05-06
Release v0.38.0

### LINKED ISSUE #196: stdout and stderr mixed in secrethub run

florisvdg:
### Added

- Add `--no-newline`, `-n` flag to `read` command to not print newline (#263)

### Fixed

- Rewrite masker to fix #196 and reduce data-dependency on timing (#267)
- Improve various help texts (#264, #271, #273)
- Fix error when autocompleting `secrethub run` command (#268)


florisvdg:
> We forgot to update this help text:
> https://github.com/secrethub/secrethub-cli/blob/ed42e92b2ffe5c5d89e675b726066012f399a4cd/internals/secrethub/run.go#L74-L76

Okay, should I just remove the last two sentences? The `masking-buffer-period` flag description is already says enough I think.

jpcoenen:
> Okay, should I just remove the last two sentences? The `masking-buffer-period` flag description is already says enough I think.

I think we should keep the part about it being a best effort

jpcoenen:
Tested new functionality on Windows and Linux. Found no problems.

florisvdg:
> I think we should keep the part about it being a best effort

Okay, what about: 

```go
const helpLong = "To protect against secrets leaking via stdout and stderr, those output streams are monitored for secrets. Detected secrets are automatically masked by replacing them with \"" + maskString + "\". " +
  "The output is buffered to scan for secrets and can be adjusted using the masking-buffer-period flag. " +
  "You should regard the masking as a best effort attempt and should always prevent secrets ending up on stdout and stderr in the first place."
```

PR #285, opened 2020-05-07
Integrated Terraform State Backend

jpcoenen:
This is an MVP of intgerating a tfstate backend into the `secrethub run` command.

It can be used by configuring the following backend in in terraform:
```hcl
terraform {
  backend "http" {
	address="http://localhost:8118"
	unlock_address="http://localhost:8118"
	lock_address="http://localhost:8118"
  	username = "path/to/state/directory"
  }
}
```

The state is served by wrapping Terraform commands in `secrethub run --tfstate`. For example:
```
secrethub run --tfstate -- terraform init -reconfigure
secrethub run --tfstate -- terraform apply
```

This opens up an HTTP-endpoint that Terraform can connect to. This HTTP-endpoint can only be connected to from processes that are a child of `secrethub run`. In other words: it should not be possible for another process than Terraform to connect to this endpoint and read/write the state.

Though not strictly necessary, it is also possible to set a password on the listener. If the secret `path/to/state/directory/password` exists, the value of it must be set in the `password` field of the Terraform backend. If this secret does not exist, the password field can be omitted.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,16 +7,19 @@ require (
 	github.com/alecthomas/kingpin v1.3.8-0.20200323085623-b6657d9477a6
 	github.com/atotto/clipboard v0.1.2
 	github.com/aws/aws-sdk-go v1.25.49
+	github.com/cakturk/go-netstat v0.0.0-20200220111822-e5b49efee7a5
 	github.com/docker/go-units v0.3.3
 	github.com/fatih/color v1.7.0
 	github.com/masterzen/winrm v0.0.0-20190308153735-1d17eaf15943
 	github.com/mattn/go-colorable v0.1.1
 	github.com/mattn/go-isatty v0.0.7
 	github.com/mitchellh/go-homedir v1.1.0
+	github.com/mitchellh/go-ps v1.0.0
 	github.com/mitchellh/mapstructure v1.1.2
 	github.com/op/go-logging v0.0.0-20160315200505-970db520ece7
 	github.com/secrethub/demo-app v0.1.0
 	github.com/secrethub/secrethub-go v0.28.0
+	github.com/weaveworks/procspy v0.0.0-20150706124340-cb970aa190c3
 	github.com/zalando/go-keyring v0.0.0-20190208082241-fbe81aec3a07
 	golang.org/x/crypto v0.0.0-20190313024323-a1f597ede03a
 	golang.org/x/sys v0.0.0-20190222072716-a9d3bda3a223
--- a/go.sum
+++ b/go.sum
@@ -20,6 +20,8 @@ github.com/aws/aws-sdk-go v1.19.38 h1:WKjobgPO4Ua1ww2NJJl2/zQNreUZxvqmEzwMlRjjm9
 github.com/aws/aws-sdk-go v1.19.38/go.mod h1:KmX6BPdI08NWTb3/sm4ZGu5ShLoqVDhKgpiN924inxo=
 github.com/aws/aws-sdk-go v1.25.49 h1:j5R2Ey+g8qaiy2NJ9iH+KWzDWS4SjXRCjhc22EeQVE4=
 github.com/aws/aws-sdk-go v1.25.49/go.mod h1:KmX6BPdI08NWTb3/sm4ZGu5ShLoqVDhKgpiN924inxo=
+github.com/cakturk/go-netstat v0.0.0-20200220111822-e5b49efee7a5 h1:BjkPE3785EwPhhyuFkbINB+2a1xATwk8SNDWnJiD41g=
+github.com/cakturk/go-netstat v0.0.0-20200220111822-e5b49efee7a5/go.mod h1:jtAfVaU/2cu1+wdSRPWE2c1N2qeAA3K4RH9pYgqwets=
 github.com/danieljoos/wincred v1.0.2 h1:zf4bhty2iLuwgjgpraD2E9UbvO+fe54XXGJbOwe23fU=
 github.com/danieljoos/wincred v1.0.2/go.mod h1:SnuYRW9lp1oJrZX/dXJqr0cPK5gYXqx3EJbmjhLdK9U=
 github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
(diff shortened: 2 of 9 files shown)


SimonBarendse:
We'd love to get your feedback on this!

If you don't want to build the binary yourself, here they are for your convenience:
[secrethub-linux-amd64.tar.gz](https://github.com/secrethub/secrethub-cli/files/4633211/secrethub-linux-amd64.tar.gz)
[secrethub-darwin-amd64.tar.gz](https://github.com/secrethub/secrethub-cli/files/4633205/secrethub-darwin-amd64.tar.gz)
[secrethub-windows-amd64.zip](https://github.com/secrethub/secrethub-cli/files/4633218/secrethub-windows-amd64.zip)

internals/integrations/tfstate/backend_other.go
```diff
@@ -0,0 +1,185 @@
...
+
+	secret, err := b.client.Secrets().ReadString(passwordPath)
+	if err != nil && !api.IsErrNotFound(err) {
+		return nil, err
+	} else if err == nil {
+		if password == "" {
+			return b.respondError(http.StatusUnauthorized, "password stored at %s should be set as auth password", passwordPath), nil
+		}
+		if password != secret {
+			return b.respondError(http.StatusForbidden, "provided password does not password stored at %s", passwordPath), nil
```
antonbabenko, line 109: `provided password does not password stored at %s` => `provided password does not match the password stored at %s`

antonbabenko:
Here is my feedback when I try to use this backend:
1. Setting username to valid username makes sense to me (`username       = "antonbabenko"`)
```
[SecretHub]: set user to a valid repository or directory, got: antonbabenko
```
The error message is confusing.

2. I set `username` to a correct value (`<namespace>/<repo>`) and it works. I see that we can't rename `username` into `path`, then there should be clear instructions on what is expected as a value.

3. When running `secrethub run --tfstate -- terraform apply` with missconfigured `username` (eg, `username       = "antonbabenko/demo/terraform.tfstate"` and this key already contains previous tfstate. Terraform apply fails like this:

```
[SecretHub]: Directory not found (server.dir_not_found)
Failed to save state: HTTP error: 404

Error: Failed to persist state to backend.

The error shown above has prevented Terraform from writing the updated state
to the configured backend. To allow for recovery, the state has been written
to the file "errored.tfstate" in the current working directory.

Running "terraform apply" again at this point will create a forked state,
making it harder to recover.

To retry writing this state, use the following command:
    terraform state push errored.tfstate
```

When trying to overwrite existing key with the one which was produced and stored locally (`errored.tfstate`), I run into another issue:

```
$ secrethub run --tfstate -- terraform state push errored.tfstate
Acquiring state lock. This may take a few moments...
[SecretHub]: Directory not found (server.dir_not_found)
Error locking state: Error acquiring the state lock: Unexpected HTTP response code 404

Terraform acquires a state lock to protect the state from being written
by multiple users at the same time. Please resolve the issue above and try
again. For most commands, you can disable locking with the "-lock=false"
flag, but this is not recommended.
```

As a solution for this, maybe `secrethub` CLI should do basic verification that key is readable and/or correct?


5. Currently, each state should be placed in separate repo or directory to prevent accidental overrides, which is very easy to forget. I don't see a way to customize the name in [HTTP backend](https://github.com/hashicorp/terraform/blob/master/backend/remote-state/http/backend.go). Do you have any ideas on how this can be improved for the users?

6. When trying to access the running process from separate execution the error message (`[SecretHub]: can only be reached from a process spawned with secrethub run`) is being printed into the main window where prompt for the confirmation is being expected. It would be better to raise the same error in the window where this belongs to (after `Error: Error loading state: HTTP remote state endpoint invalid auth`).

That's it for now :) Let me know when an update is coming.

internals/integrations/tfstate/backend_other.go
jpcoenen: Good catch :ok_hand:

jpcoenen:
Thanks a lot for the feedback @antonbabenko! Seems like you did some really thorough testing :grinning: 

> Setting username to valid username makes sense to me (username = "antonbabenko")
> [SecretHub]: set user to a valid repository or directory, got: antonbabenko
> The error message is confusing.

I think you already figured out that we're a little boun to what we can do with the HTTP State Backend; that's why the `username` field is used. But I do definitely agree that this error message could be way more helpful. I'll work on that.

> When running secrethub run --tfstate -- terraform apply with missconfigured username (eg, username = "antonbabenko/demo/terraform.tfstate" and this key already contains previous tfstate. Terraform apply fails like this:

You definitely caught a bug here. I think it goes wrong because a `404` is returned, which Terraform interprets as "all is correct, but there is no state saved yet", but it should instead return an error that signals that something failed. That sounds fixable. 

I'll also make the error message itself more helpful than just `Directory not found (server.dir_not_found)`.

> Currently, each state should be placed in separate repo or directory to prevent accidental overrides, which is very easy to forget. I don't see a way to customize the name in HTTP backend. Do you have any ideas on how this can be improved for the users?

Hmm. That's an interesting question. The reason for using a directory as input instead of the path to the secret where the state is stored, is that this directory is both used to store the state and the lock. I do not directly see a way around that. 

What would happen if you used the same S3 path for 2 different Terraform projects? Is that different from how this works? Or is it mainly that there is some confusion from the `username` being used to specify the path where the state is stored?

> When trying to access the running process from separate execution the error message ([SecretHub]: can only be reached from a process spawned with secrethub run) is being printed into the main window where prompt for the confirmation is being expected. It would be better to raise the same error in the window where this belongs to (after Error: Error loading state: HTTP remote state endpoint invalid auth).

You mean that it would be better to also print `can only be reached from a process spawned with secrethub run` in the terminal where you execute the failing `terraform` command? Because I agree that would be helpful, but it's not possible, unfortunately. Terraform does not allow any customization of the errors it displays. The messages prepended with `[SecretHub` are output by the SecretHub CLI, not by Terraform.

Or.. are you running `secrethub run -- terraform apply` in two terminals simultaneously in this case? Because in that case, it would be possible to make some improvements.

--- 

To conclude: really great feedback! I'll get to work on the above points in the near future. Though I cannot precisely say when, I will let you know once a new iteration is done.

One important question that's still on my mind: if we do some iterations (starting with the points you mentioned), how usable would you find it?

jpcoenen:
@antonbabenko I've just updated the error message you describe in point 1 and have introduced an explicit check for the existence of the given directory, which should help with point 3.

Here is the new build:
[secrethub-linux-amd64.tar.gz](https://github.com/secrethub/secrethub-cli/files/4786303/secrethub-linux-amd64.tar.gz)
[secrethub-darwin-amd64.tar.gz](https://github.com/secrethub/secrethub-cli/files/4786305/secrethub-darwin-amd64.tar.gz)
[secrethub-windows-amd64.zip](https://github.com/secrethub/secrethub-cli/files/4786306/secrethub-windows-amd64.zip)

antonbabenko:
Nice! I will try to take a look at this release and comment during next week.

PR #287, opened 2020-05-20
Update readme

jpcoenen:
Changes include but not limited to:
- Divided usage examples in three different sections
- More usage examples
- Added integrations sections
- Changed description of "SecretHub"
- Shortened the Get Started section to only refer to the Getting Started Guide

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,9 @@ Revoke an account's access to a repository.
 
 SecretHub integrates with all the tools you already know and love.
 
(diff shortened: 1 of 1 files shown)


mackenbach:
Sweet! Very well done. One low-hanging fruit thing could also be: two buttons below the logo at the top, like we have at https://github.com/secrethub/secrethub-circleci-orb 

The two buttons I'd suggest to put there are `Get started` and `View docs`. Non-blocking improvement though.

PR #290, opened 2020-05-30
Support multiple arguments for mkdir command

Marton6:
This PR adds support for creating multiple directories with the `mkdir` command by passing multiple arguments to it, as well as tests for verifying this functionality.


Marton6:
Yes, that's a good idea. I'll implement this behavior.

internals/secrethub/mkdir.go
```diff
@@ -47,40 +49,41 @@ func (cmd *MkDirCommand) Run() error {
...
 
 	for _, path := range cmd.paths {
-		if cmd.parents {
-			err = client.Dirs().CreateAll(path.Value())
-			if err != nil {
-				return err
-			}
+		err := cmd.createDirectory(client, path)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Encountered an error: %s\n", err)
```
jpcoenen, line 54: I'd expect this error to be more in line with the success message.  Now you can get the following output:
```
Encountered an error: Directory or secret already exists, create a different directory (server.dir_already_exists) 
Encountered an error: Directory or secret already exists, create a different directory (server.dir_already_exists) 
Created a new directory at joriscoenen/dirtest/staging/service4
```
This leaves me to guess which directories resulted in an error. 

For example:
```suggestion
fmt.Fprintf(os.Stderr, "Could not create a new directory at %s: %s\n", path, err)
```

jpcoenen:
We're not there yet. Now the message is really confusing when using `--parents` and the directory already exists (it always kind of has been confusing). It will tell you multiple times that it has created a new directory, which is not the case.

jpcoenen:
Hmm. In the way `secrethub-go` handles `Dirs().CreateAll()`, this is not easy to fix. So we'd probably leave it as is for now. One thing is that it's not that different from the status quo.

PR #293, opened 2020-06-04
Make proxy for HTTP-client configurable through flag

jpcoenen:
Allows a user to set the `SECRETHUB_PROXY_ADDRESS` environment variable or `--proxy-address` flag to configure a proxy used when connecting to the API. It allows for more precise control than just using the `HTTP_PROXY` environment variable because this only affects the SecretHub CLI.

### DIFF
--- a/internals/secrethub/client_factory.go
+++ b/internals/secrethub/client_factory.go
@@ -63,14 +63,6 @@ func (f *clientFactory) NewClient() (secrethub.ClientInterface, error) {
 		options := f.baseClientOptions()
 		options = append(options, secrethub.WithCredentials(credentialProvider))
 
-		if f.proxyAddress != nil {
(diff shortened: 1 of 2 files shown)


florisvdg:
A test would be nice (also testing environment variable precedence, `SECRETHUB_HTTP_ADDRESS` > `HTTP_PROXY`).

jpcoenen:
> A test would be nice (also testing environment variable precedence, `SECRETHUB_PROXY_ADDRESS` > `HTTP_PROXY`).

I realized the same after I pushed. There's a simple test in there now. Checks whether a connection is made to the proxy server. I leave the testing of actual proxying to the maintainers of Go.

mackenbach:
Yes gotta love the elegance of adding an option like this. So smooth 😍 

And +1 on Simon's idea of moving this to the client at some point.

PR #298, opened 2020-06-23
GCP linking

jpcoenen:


internals/secrethub/service_gcp_link.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		return err
+	}
+
+	_, err = client.IDPLinks().GCP().Get(cmd.namespace, cmd.projectID)
+	if api.IsErrNotFound(err) {
+		return err
+	}
+
+	question := fmt.Sprintf("Are you sure you want to delete the link link between GCP project %s and the namespace %s? This does not affect any existing service accounts.", cmd.projectID, cmd.namespace)
```
florisvdg, line 139: Could be helpful to describe the consequence of deleting the link: 
```suggestion
	question := fmt.Sprintf("Are you sure you want to delete the link link between GCP project %s and the namespace %s? Without the link, you cannot create new service accounts for this GCP project. This does not affect existing service accounts.", cmd.projectID, cmd.namespace)
```

internals/secrethub/service_gcp_init.go
```diff
@@ -108,6 +109,31 @@ func (cmd *ServiceGCPInitCommand) Run() error {
...
 	}
 
+	projectID, err := api.ProjectIDFromGCPEmail(cmd.serviceAccountEmail)
+	if err != nil {
+		return fmt.Errorf("invalid service account email: %s", err)
+	}
+
+	_, err = client.IDPLinks().GCP().Get(cmd.repo.GetNamespace(), projectID)
+	if api.IsErrNotFound(err) {
+		fmt.Fprintf(cmd.io.Output(), "GCP project %s is not yet linked to the namespace %s. ", projectID, cmd.repo.GetNamespace())
```
SimonBarendse, line 119: Do you think we should explain what it means for a project to be linked to the namespace and what are the consequences?

For example: Is this a one-on-one link, can I still create other service accounts for this project in other namespaces?

I think we want to explain to the user why they want to answer yes in most cases, but also, why they might want to answer no in other cases.

internals/secrethub/service_gcp_init.go
```diff
@@ -108,6 +109,31 @@ func (cmd *ServiceGCPInitCommand) Run() error {
...
+
+		confirm, err := ui.AskYesNo(cmd.io, "Do you want to create this link now?", ui.DefaultYes)
```
SimonBarendse, line 121: Do we want to check whether users can perform this action first?

SimonBarendse: And in the case they cannot: print that they should ask an admin to run `secrethub service gcp link <namespace> <project-id>`

SimonBarendse: In a future version it would be nice if they can sent a request for this to the admins using SecretHub, for now printing will suffice I think.

internals/secrethub/service_gcp_link.go
```diff
@@ -0,0 +1,193 @@
...
+	if !api.IsErrNotFound(err) {
+		fmt.Fprintf(cmd.io.Output(), "Namespace %s and GCP project %s are already linked.\n", cmd.namespace, cmd.projectID)
+		return nil
+	}
+
+	return createGCPLink(client, cmd.io, cmd.namespace, cmd.projectID)
+}
+
+func (cmd *ServiceGCPLinkCommand) Register(r command.Registerer) {
+	clause := r.Command("link", "Create a new link between a namespace and a GCP project to allow creating SecretHub service accounts for GCP Service Accounts in the GCP project.")
```
SimonBarendse, line 48: Maybe we can explain in a `clause.HelpLong` why a link is required?

SimonBarendse: And shall we add that a GCP project can be linked to multiple namespaces and that a namespace can be linked to multiple GCP projects?

internals/secrethub/service_gcp_link.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+func (cmd *ServiceGCPLinkCommand) Run() error {
+	client, err := cmd.newClient()
+	if err != nil {
+		return err
+	}
+
+	_, err = client.IDPLinks().GCP().Get(cmd.namespace, cmd.projectID)
+	if !api.IsErrNotFound(err) {
```
SimonBarendse, line 39: How would you feel about adding an `Exists` func to `client.IDPLinks().GCP()` which does exactly this?

SimonBarendse: This could also be used in the delete command.

internals/secrethub/service_gcp_init.go
jpcoenen: You're right that for the user-experience that would be best. But.. we currently do not have a mechanism in place for checking permissions. I can imagine these two relatively easy methods of checking:
1. Try the request with an invalid authorization code. If it results in a 403, we know the user cannot do this action.
2. Check if the user has admin permission on the repo. Does work now, but would lead to problems if we ever change anything in the permission model.

Both are not ideal imo. Did you have any other ideas?

internals/secrethub/service_gcp_link.go
jpcoenen: Added quite a long `HelpLong`

internals/secrethub/service_gcp_link.go
jpcoenen: :check

internals/secrethub/service_gcp_init.go
jpcoenen: Added some more text

jpcoenen:
@florisvdg I changed the start of the linking process to what we discussed: first explain the process and then ask for input to continue. What do you think?

internals/secrethub/service_gcp_link.go
```diff
@@ -53,6 +53,21 @@ func (cmd *ServiceGCPLinkCommand) Register(r command.Registerer) {
 	clause.Arg("namespace", "The SecretHub namespace to link.").Required().SetValue(&cmd.namespace)
 	clause.Arg("project-id", "The GCP project to link the namespace to.").Required().SetValue(&cmd.projectID)
 
+	clause.HelpLong("Linking a GCP project to a namespace is required to create SecretHub service accounts that use a GCP Service Account within the project. " +
+		"A SecretHub namespace can be linked to multiple GCP projects and a GCP project can be linked to multiple namespaces.\n" +
+		"\n" +
+		"As long as the link exists, new service accounts can be created for the GCP project. " +
+		"If a link is deleted, no new services can be created, but previously created services are unaffected. \n" +
+		"\n" +
+		"This command will open browser where you are asked to authorize SecretHub to perform iam.test on your GCP projects. " +
```
SimonBarendse, line 62:
```suggestion
		"This command will open your browser where you are asked to authorize SecretHub to perform iam.test on your GCP projects. " +
```

Alternatives: `a` or `the`

internals/secrethub/service_gcp_link.go
```diff
@@ -53,6 +53,21 @@ func (cmd *ServiceGCPLinkCommand) Register(r command.Registerer) {
...
+		"This authorization is used to verify that you have access to the specified GCP project. " +
+		"It is therefore important that the GCP account selected during the authorization process has access to the GCP project.\n" +
+		"\n" +
+		"Once the granted authorization has been used to confirm your access to the GCP project, the authorization will automatically be revoked. " +
+		"This can be verified by going to https://myaccount.google.com/permissions. " +
```
SimonBarendse, line 67: Nice! :ok_hand:

internals/secrethub/service_gcp_link.go
SimonBarendse: Sweet! Really nice

PR #300, opened 2020-07-01
Move all app construction to secrethub.NewApp function

SimonBarendse:


florisvdg:
One thing though: this change by itself is not enough to expose the Kingpin model: `app.cli` is still private.

SimonBarendse:
> One thing though: this change by itself is not enough to expose the Kingpin model: app.cli is still private.

Seems like I've overlooked that one. We can either create a wrapper for the `Model` function or make `app.cli` public. I'm fine with both options. Maybe we can even embed `app.cli` instead of having it as a field?

PR #301, opened 2020-07-02
Run linter in Docker

SimonBarendse:
This makes it possible to have different linter versions locally and makes sure "make lint" always runs with the version specified in the Makefile.


SimonBarendse:
Haha jup, basically we don't want any of the environment separation except for the binary version. :sweat_smile:

PR #303, opened 2020-07-02
Increase clipboard timeout to 3 minutes

SimonBarendse:
45 seconds might not be enough time to paste the value copied to the clipboard. Giving the user a little more time is desirable, while still clearing the clipboard within a reasonable time to prevent the secret sprawling around.

### DIFF
--- a/internals/secrethub/clear_clipboard.go
+++ b/internals/secrethub/clear_clipboard.go
@@ -12,7 +12,7 @@ import (
 )
 
 // defaultClearClipboardAfter defines the default TTL for data written to the clipboard.
-const defaultClearClipboardAfter = 45 * time.Second
+const defaultClearClipboardAfter = 3 * time.Minute
 
 // ClearClipboardCommand is a command to clear the contents of the clipboard after some time passed.
 type ClearClipboardCommand struct {
--- a/internals/secrethub/service_init.go
+++ b/internals/secrethub/service_init.go
@@ -88,7 +88,7 @@ func (cmd *ServiceInitCommand) Run() error {
 			return err
 		}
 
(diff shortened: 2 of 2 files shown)


stavalfi:
where is the repo of the docs?

internals/secrethub/clear_clipboard.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 // defaultClearClipboardAfter defines the default TTL for data written to the clipboard.
-const defaultClearClipboardAfter = 45 * time.Second
+const defaultClearClipboardAfter = 3 * time.Minute
```
jpcoenen, line 15: Is this also necessary for `read`?

SimonBarendse:
> where is the repo of the docs?

Hi @stavalfi , the CLI reference docs are still closed-source unfortunately. Eventually, we want to generate the docs based on the help text that's defined on the commands in this open-source repo (we're not there yet though).

Is there a specific reason you're looking for the docs repo?

stavalfi:
I couldn't find a way to search free text in the docs website. github provides a way to search inside a repo

internals/secrethub/clear_clipboard.go
jpcoenen: We discussed this in person: increasing the timeout for a might reduce the chance a little bit that the clipboard is cleared before using it, but at the same time, it also increases the chance of an accidental paste in the wrong location. Since using `--clip` for a read is a conscious decision, the fact that the clipboard is cleared is printed and it can easily be repeated by re-running the command, there is no real need to drastically increase this timeout (or at least not this much). 

---

For initializing services this is another story because this is not 100% repeatable; if the key gets cleared from the clipboard, it can never be retrieved and the service account must be recreated. So we could increase the timeout here, but it might be better to solve the underlying issue itself. Because the question is whether it is solved when we increase the timeout to 3 minutes.

PR #308, opened 2020-07-06
Rephrase GCP service init help-text

SimonBarendse:
Consitently talk about "SecretHub GCP service accounts" and don't mix this with "service accounts that use the native GCP identity provider".

We don't explain what we mean by native identity provider in this help text, so using that term here might be confusing. I believe that sticking to the simpler "service account" might create a more consistent and therefore easier to understand help text.


SimonBarendse:
We'll circle back to this later and pick this up more holistically throughout all documentation.

PR #310, opened 2020-07-06
Directly check if repo exists

jpcoenen:
This avoids an annoying error at the end if the repo does not exist.

### DIFF
--- a/internals/secrethub/service_gcp_init.go
+++ b/internals/secrethub/service_gcp_init.go
@@ -48,6 +48,11 @@ func (cmd *ServiceGCPInitCommand) Run() error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


internals/secrethub/service_gcp_init.go
```diff
@@ -48,6 +48,11 @@ func (cmd *ServiceGCPInitCommand) Run() error {
 		return err
 	}
 
+	_, err = client.Repos().Get(cmd.repo.String())
```
SimonBarendse, line 51: I think this would benefit from a comment.
```suggestion
	// Fail fast if the repo does not exist.
	_, err = client.Repos().Get(cmd.repo.String())
```

PR #311, opened 2020-07-07
Do not accept a directory as argument to service init

SimonBarendse:
Just like we already do for `service aws init`, don't accept a directory as argument of the command. Instead, if you want to set a permission on a subdirectory, you should use the `--permission` flag.


jpcoenen:
Did you actually test this? Because it did not build.

SimonBarendse:
I have not.

PR #313, opened 2020-07-07
Only list user-managed GCP Service Accounts in service gcp init prompt

jpcoenen:
Other Service Accounts are not supported anyway, so no reason to list them.

### DIFF
--- a/internals/secrethub/service_gcp_init.go
+++ b/internals/secrethub/service_gcp_init.go
@@ -68,7 +68,7 @@ func (cmd *ServiceGCPInitCommand) Run() error {
 		serviceAccountLister := gcpServiceAccountOptionLister{
 			ProjectID: projectID,
 		}
-		serviceAccountEmail, err := ui.ChooseDynamicOptionsValidate(cmd.io, "What is the email of the service account you want to use?", serviceAccountLister.Options, "service account", api.ValidateGCPServiceAccountEmail)
+		serviceAccountEmail, err := ui.ChooseDynamicOptionsValidate(cmd.io, "What is the email of the service account you want to use?", serviceAccountLister.Options, "service account", api.ValidateGCPUserManagedServiceAccountEmail)
 		if err != nil {
 			return err
 		}
@@ -90,7 +90,7 @@ func (cmd *ServiceGCPInitCommand) Run() error {
 	}
 
 	if cmd.serviceAccountEmail == "" {
-		serviceAccountEmail, err := ui.AskAndValidate(cmd.io, "What is the email of the GCP Service Account that should have access to the service?\n", 3, api.ValidateGCPServiceAccountEmail)
+		serviceAccountEmail, err := ui.AskAndValidate(cmd.io, "What is the email of the GCP Service Account that should have access to the service?\n", 3, api.ValidateGCPUserManagedServiceAccountEmail)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


internals/secrethub/service_gcp_init.go
```diff
@@ -216,16 +216,19 @@ func (l *gcpServiceAccountOptionLister) Options() ([]ui.Option, bool, error) {
 		return nil, false, gcp.HandleError(err)
 	}
 
-	options := make([]ui.Option, len(resp.Accounts))
-	for i, account := range resp.Accounts {
+	options := make([]ui.Option, 0, len(resp.Accounts))
+	for _, account := range resp.Accounts {
+		if err := api.ValidateGCPServiceAccountEmail(account.Email); err != nil {
```
SimonBarendse, line 221: From the commit message I get that there are GCP service accounts which do not pass this validation and that those are not "user managed". This is not clear to me from reading the code. Maybe we can rename the `ValidateGCPServiceAccountEmail` function so that it reflects that it validates user managed service accounts and/or maybe we can add a comment above this call and/or to `ValidateGCPServiceAccountEmail` that explain the difference between these two types of service accounts and which is included in this validation?

florisvdg: As for the naming, how it validates is implementation detail, so the name is fine by me. But a comment here would be good indeed:

```suggestion
		// Only list user-managed service accounts
		if err := api.ValidateGCPServiceAccountEmail(account.Email); err != nil {
```

SimonBarendse: I agree that how it validates is implementation detail, I believe it is important that the name conveys _what_ is validated. Does the current name do so? Are service accounts that are not "user managed" not represented by an email address?

florisvdg: Suggestion above good enough?

SimonBarendse: Maybe even better, we can create a small wrapper function that explains this:

```go
// isUserManagedGCPServiceAccount checks whether a GCP service account is user managed.
// some more explanation what it means to be "user managed"
func isUserManagedGCPServiceAccount(account iam.ServiceAccount) bool {
  // Only user-managed service accounts have an email address?
  err := api.ValidateGCPServiceAccountEmail(account.Email)
  return err != nil
}
```

Then we can do:

```go
if isUserManagedGCPServiceAccount(account) {
  continue
}
```

SimonBarendse: > Suggestion above good enough?

Well, to me it doesn't make it clear what is going on. From what I understand, this piece of code works because service accounts which are not "user managed" don't have a (valid?) email address set. It would help if that is (explicitly) said in the comment.

What's more, it's also not really clear to be what is a "user managed" service account. I think it would be helpful if this is documented as well, unless this is a term GCP is using as well.

SimonBarendse: Ah, GCP uses the concept of [service accounts with user-managed keys](https://cloud.google.com/iam/docs/service-accounts#user-managed_keys). I think it's clear enough that we mean those.

SimonBarendse: Gave the validation function a closer look. It appears to do validate email addresses for user-managed service accounts specifically: https://github.com/secrethub/secrethub-go/blob/08657e1d8043614652891f7328370755b37d890b/internals/api/patterns.go#L314

I believe the name should reflect this.

jpcoenen: > Ah, GCP uses the concept of [service accounts with user-managed keys](https://cloud.google.com/iam/docs/service-accounts#user-managed_keys). I think it's clear enough that we mean those.

No, it literally means ["user managed service accounts"](https://cloud.google.com/iam/docs/service-accounts#user-managed), in contrary to Google Managed Service Accounts

jpcoenen: Changed it in `secrethub-go`: https://github.com/secrethub/secrethub-go/pull/202

PR #319, opened 2020-07-20
Add `-f`, `-i` and `--noreport` flags to tree command

edif2008:
This PR contains the implementation of 3 flags for the `tree` command in the CLI:
 - `-f` for full path of the directories and secrets
 - `-i` for no indentation
 - `--no-report` for skipping the report at the end

### DIFF
--- a/internals/secrethub/tree.go
+++ b/internals/secrethub/tree.go
@@ -109,10 +109,10 @@ func (cmd *TreeCommand) printDirContentsRecursively(dir *api.Dir, prefix string,
 
 		if i == total-1 {
 			fmt.Fprintf(w, format[0], prefix, prevPath, name)
-			cmd.printDirContentsRecursively(sub, prefix+format[2], w, format, prevPath + name.(string))
+			cmd.printDirContentsRecursively(sub, prefix+format[2], w, format, prevPath+name.(string))
 		} else {
 			fmt.Fprintf(w, format[1], prefix, prevPath, name)
-			cmd.printDirContentsRecursively(sub, prefix+format[3], w, format, prevPath + name.(string))
+			cmd.printDirContentsRecursively(sub, prefix+format[3], w, format, prevPath+name.(string))
 		}
 		i++
 	}
--- a/internals/secrethub/tree.go
+++ b/internals/secrethub/tree.go
@@ -53,13 +53,21 @@ func (cmd *TreeCommand) Register(r command.Registerer) {
 	clause.Flag("full-paths", "Print the full paths of the directories and secrets.").Short('f').BoolVar(&cmd.fullPaths)
 	clause.Flag("no-indentation", "Print the content without indentation.").Short('i').BoolVar(&cmd.noIndentation)
 	clause.Flag("no-report", "Skip the report at the bottom.").BoolVar(&cmd.noReport)
+	clause.Flag("noreport", "Skip the report at the bottom.").BoolVar(&cmd.noReport)
 
 	command.BindAction(clause, cmd.Run)
 }
 
 // printTree recursively prints the tree's contents in a tree-like structure.
 func (cmd *TreeCommand) printTree(t *api.Tree, w io.Writer) {
-	name := colorizeByStatus(t.RootDir.Status, t.RootDir.Name)
+
+	rootDirName := func() string {
+		if cmd.fullPaths {
+			return cmd.path.Value()
+		}
+		return t.RootDir.Name
+	}()
+	name := colorizeByStatus(t.RootDir.Status, rootDirName)
 	fmt.Fprintf(w, "%s\n", name)
 
 	var format [4]string
@@ -75,7 +83,7 @@ func (cmd *TreeCommand) printTree(t *api.Tree, w io.Writer) {
 		format[3] = "│   "
 	}
 	if cmd.fullPaths {
-		cmd.printDirContentsRecursively(t.RootDir, "", w, format, t.RootDir.Name)
+		cmd.printDirContentsRecursively(t.RootDir, "", w, format, cmd.path.Value())
 	} else {
 		cmd.printDirContentsRecursively(t.RootDir, "", w, format, "")
 	}
@@ -97,33 +105,37 @@ func (cmd *TreeCommand) printDirContentsRecursively(dir *api.Dir, prefix string,
 
 	total := len(dir.SubDirs) + len(dir.Secrets)
 
-	if !cmd.fullPaths {
-		prevPath = ""
-	} else {
+	if cmd.fullPaths {
 		prevPath += "/"
+	} else {
+		prevPath = ""
 	}
 
 	i := 0
 	for _, sub := range dir.SubDirs {
-		name := colorizeByStatus(sub.Status, sub.Name)
+		colorName := colorizeByStatus(sub.Status, sub.Name)
+		colorPrefix := colorizeByStatus(sub.Status, prefix)
+		colorPrevPath := colorizeByStatus(sub.Status, prevPath)
 
 		if i == total-1 {
-			fmt.Fprintf(w, format[0], prefix, prevPath, name)
-			cmd.printDirContentsRecursively(sub, prefix+format[2], w, format, prevPath+name.(string))
+			fmt.Fprintf(w, format[0], colorPrefix, colorPrevPath, colorName)
+			cmd.printDirContentsRecursively(sub, prefix+format[2], w, format, prevPath+sub.Name)
 		} else {
-			fmt.Fprintf(w, format[1], prefix, prevPath, name)
-			cmd.printDirContentsRecursively(sub, prefix+format[3], w, format, prevPath+name.(string))
+			fmt.Fprintf(w, format[1], colorPrefix, colorPrevPath, colorName)
+			cmd.printDirContentsRecursively(sub, prefix+format[3], w, format, prevPath+sub.Name)
 		}
 		i++
 	}
 
 	for _, secret := range dir.Secrets {
-		name := colorizeByStatus(secret.Status, secret.Name)
+		colorName := colorizeByStatus(secret.Status, secret.Name)
+		colorPrefix := colorizeByStatus(secret.Status, prefix)
+		colorPrevPath := colorizeByStatus(secret.Status, prevPath)
 
 		if i == total-1 {
-			fmt.Fprintf(w, format[0], prefix, prevPath, name)
+			fmt.Fprintf(w, format[0], colorPrefix, colorPrevPath, colorName)
 		} else {
-			fmt.Fprintf(w, format[1], prefix, prevPath, name)
+			fmt.Fprintf(w, format[1], colorPrefix, colorPrevPath, colorName)
 		}
 		i++
 	}
--- a/internals/secrethub/tree.go
+++ b/internals/secrethub/tree.go
@@ -88,12 +88,6 @@ func (cmd *TreeCommand) printTree(t *api.Tree, w io.Writer) {
 // in a tree-like structure, subdirs first followed by secrets.
 func (cmd *TreeCommand) printDirContentsRecursively(dir *api.Dir, prefix string, w io.Writer, prevPath string) {
 
-	const (
-		format1 = "%s\n"
-		format2 = "%s└── %s\n"
-		format3 = "%s├── %s\n"
-	)
-
 	sort.Sort(api.SortDirByName(dir.SubDirs))
 	sort.Sort(api.SortSecretByName(dir.Secrets))
 
@@ -115,13 +109,13 @@ func (cmd *TreeCommand) printDirContentsRecursively(dir *api.Dir, prefix string,
 		colorName := colorizeByStatus(sub.Status, name+"/")
 
 		if cmd.noIndentation {
-			fmt.Fprintf(w, format1, colorName)
+			fmt.Fprintf(w, "%s\n", colorName)
 			cmd.printDirContentsRecursively(sub, prefix, w, name)
 		} else if i == total-1 {
-			fmt.Fprintf(w, format2, prefix, colorName)
+			fmt.Fprintf(w, "%s└── %s\n", prefix, colorName)
 			cmd.printDirContentsRecursively(sub, prefix+"    ", w, name)
 		} else {
-			fmt.Fprintf(w, format3, prefix, colorName)
+			fmt.Fprintf(w, "%s├── %s\n", prefix, colorName)
 			cmd.printDirContentsRecursively(sub, prefix+"│   ", w, name)
 		}
 		i++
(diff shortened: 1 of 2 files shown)


internals/secrethub/tree.go
```diff
@@ -50,26 +50,44 @@ func (cmd *TreeCommand) Register(r command.Registerer) {
 	clause := r.Command("tree", "List contents of a directory in a tree-like format.")
 	clause.Arg("dir-path", "The path to to show contents for").Required().PlaceHolder(optionalDirPathPlaceHolder).SetValue(&cmd.path)
 
+	clause.Flag("full-paths", "Print the full paths of the directories and secrets.").Short('f').BoolVar(&cmd.fullPaths)
+	clause.Flag("no-indentation", "Print the content without indentation.").Short('i').BoolVar(&cmd.noIndentation)
+	clause.Flag("no-report", "Skip the report at the bottom.").BoolVar(&cmd.noReport)
```
SimonBarendse, line 55: The Unix tree command has a `--noreport` flag (without a dash). Shall we rename this one or accept both alternatives?

I believe there's no native flag alias support in Kingpin, but we could register two flags?

internals/secrethub/tree.go
```diff
@@ -81,11 +99,11 @@ func printDirContentsRecursively(dir *api.Dir, prefix string, w io.Writer) {
 		name := colorizeByStatus(sub.Status, sub.Name)
 
 		if i == total-1 {
-			fmt.Fprintf(w, "%s└── %s/\n", prefix, name)
-			printDirContentsRecursively(sub, prefix+"    ", w)
+			fmt.Fprintf(w, format[0], prefix, name)
```
SimonBarendse: I'm struggling a bit to read from the `format` variable what is going to be printed in this case (it's defined in another function, `printTree`). How would you feel about defining the format in this function? One way I can imagine is the following:

```Golang
if cmd.noIndentation {
    fmt.Print(w, name)
} else if i == total-1 {
    fmt.Fprintf(w, "%s└── %s/\n", prefix, name)
    prefix += "    "
} else {
    fmt.Fprintf(w, "%s├── %s/\n", prefix, name)
    prefix += "│   "
}
printDirContentsRecursively(sub, prefix, w)
```

And maybe we can put this in a separate function so that we can re-use the same thing for directories and secrets? It looks like they both use the same format in the same cases.

internals/secrethub/tree.go
```diff
@@ -87,23 +90,29 @@ func (cmd *TreeCommand) printTree(t *api.Tree, w io.Writer) {
...
 	sort.Sort(api.SortDirByName(dir.SubDirs))
 	sort.Sort(api.SortSecretByName(dir.Secrets))
 
 	total := len(dir.SubDirs) + len(dir.Secrets)
 
+	if !cmd.fullPaths {
+		prevPath = ""
+	} else {
+		prevPath += "/"
+	}
```
SimonBarendse, line 104: Nitpicky level high:

In case where we're dealing with both the truethy and the falsy value of a variable (if/else), I usually find it more readable not the use negation, to prevent the else case from being a negation of a negation (not not full paths).

How would you feel about changing to the following?

```suggestion
	if cmd.fullPaths {
		prevPath += "/"
	} else {
		prevPath = ""
	}
```

internals/secrethub/tree.go
```diff
@@ -47,42 +50,69 @@ func (cmd *TreeCommand) Register(r command.Registerer) {
...
+		format[2] = ""
+		format[3] = ""
+	} else {
+		format[0] = "%s└── %s%s\n"
+		format[1] = "%s├── %s%s\n"
+		format[2] = "    "
+		format[3] = "│   "
+	}
+	if cmd.fullPaths {
+		cmd.printDirContentsRecursively(t.RootDir, "", w, format, t.RootDir.Name)
```
SimonBarendse, line 78: To print the full path, I believe we need to pass the full path of the root directory, instead of the name:

```suggestion
		cmd.printDirContentsRecursively(t.RootDir, "", w, format, cmd.path.Value())
```

Currently, when I do:

```
secrethub tree simon/test
```

I'm seeing this tree:
```
test
├── test/directory
│   └── test/directory/secret
└── test/s1

1 directory, 2 secrets
```

And I would expect this tree:
```
simon/test
├── simon/test/directory
│   └── simon/test/directory/secret
└── simon/test/s1

1 directory, 2 secrets
```

internals/secrethub/tree.go
```diff
@@ -47,42 +50,69 @@ func (cmd *TreeCommand) Register(r command.Registerer) {
...
+
 	i := 0
 	for _, sub := range dir.SubDirs {
 		name := colorizeByStatus(sub.Status, sub.Name)
 
 		if i == total-1 {
-			fmt.Fprintf(w, "%s└── %s/\n", prefix, name)
-			printDirContentsRecursively(sub, prefix+"    ", w)
+			fmt.Fprintf(w, format[0], prefix, prevPath, name)
+			cmd.printDirContentsRecursively(sub, prefix+format[2], w, format, prevPath+name.(string))
```
SimonBarendse, line 112: Because `name` is already colorized (made red) in case the directory is flagged, the directory part of the path of other secrets is also flagged.

What's more, in case secrets/directories are flagged, only part of the full path is printed in red.

I'll give an example to illustrate:

In this example with two secrets, only `simon/my-app/mailchimp/api-key` is flagged, `simon/my-app/stripe/api-key` is still okay. However, if I run `secrethub tree -f simon/test`, everything marked with astrixes becomes red:

    simon/**my-app**
    ├── simon/**my-app**/stripe
    │   └── simon/**my-app**/stripe/api-key
    └── simon/**my-app**/**mailchimp**
        └── simon/**my-app**/**mailchimp**/**api-key**

This is what I would expect instead:
```
    **simon/my-app**
    ├── simon/my-app/stripe
    │   └── simon/my-app/stripe/api-key
    └── **simon/my-app/mailchimp**
        └── **simon/my-app/mailchimp/api-key**
```

I believe we can achieve this by making two changes:
1. Evaluate `colorizeByStatus` on the full path instead of just the name
1. Evaluate `colarizeByStatus` only on the string passed to `fmt.Fprintf`, not on the string passed to the recursive call (`cmd.printDirContentsRecursively`)

internals/secrethub/tree.go
hculea: I think that rather the latter, such that we are compliant both to the kebab-case and to Unix's legacy naming. I'll handle that.

internals/secrethub/tree.go
hculea: Good point! Onto it.

internals/secrethub/tree.go
hculea: This is actually something that I and @edif2008 were discussing today, about how to grab the namespace of the repo! Thanks for the tip!

internals/secrethub/tree.go
hculea: This should now be fixed.

internals/secrethub/tree.go
hculea: This should now be fixed.

internals/secrethub/tree.go
hculea: This should now be fixed.

internals/secrethub/tree.go
hculea: This should now be fixed.

internals/secrethub/tree.go
hculea: I believe the reasoning for having `format` where it currently is starts from the fact that, in this way, we can avoid having to do the check for `noIndent` every time the recursive function calls itself. By having it declared before the first call of the function, we can decide beforehand which prefixes are supposed to be used. (therefore we only create 1 decision point based on `noIndent`, instead of O(n)) The alternative to this would probably be having the prefixes as global variables, but I think this is not a very good practice. Maybe @edif2008 can confirm/chime in? Let us know what you think, @SimonBarendse.

internals/secrethub/tree.go
SimonBarendse: With prefixes (sometimes) colorized, this doesn't look very good yet. Please try and run this for yourselves to see the result.

You can create a flagged secret by [creating a service account](https://secrethub.io/docs/reference/cli/service/#init), reading a secret with the service account and then [revoking the service account](https://secrethub.io/docs/reference/cli/service/#revoke).

If you have any questions, let me know :smile:

internals/secrethub/tree.go
SimonBarendse: You are right that the current code is slightly faster than the proposed changes. However, as far as I can see they both have the same runtime complexity (constant within the function + same recursive call), meaning they grow at the same pace.

What's more, the function is always called with a relatively low `n`. Therefore the function is always fast and the runtime complexity isn't too important. No human would be able to notice the difference.
Therefore, I'd pick readability over speed in this trade-off.

To get a feeling for the speed difference between the two, I've created the following benchmark:
```Golang
package main

import (
	"testing"
)

func BenchmarkWithCondition(b *testing.B) {
	// run the Fib function b.N times
	for n := 0; n < b.N; n++ {
		withCondition(true)
	}
}

func BenchmarkWithLookup(b *testing.B) {
	// run the Fib function b.N times
	for n := 0; n < b.N; n++ {
		withLookup([1]string{"foo"})
	}
}

func withCondition(condition bool) string {
	if condition {
		return "foo"
	}
	return "bar"
}

func withLookup(result [1]string) string {
	return result[0]
}
```

These were my results:
```
BenchmarkWithCondition-8   	1000000000	         0.248 ns/op
BenchmarkWithLookup-8      	1000000000	         0.245 ns/op
```

This means there's `0.000000000003 seconds` difference between evaluating one secret/directory. In a tree with a thousand secrets and directories (this is already a huge tree), this would make a difference of `0.000000003 seconds`.

hculea: Indeed! Thank you for the clarification. This should now be fixed.

internals/secrethub/tree.go
hculea: Thank you, @SimonBarendse, we now tested it on our machines and, to our knowledge, it works smoothly. Have a look and let us know if you can find any more issues.

mackenbach: > we now tested it on our machines and, to our knowledge, it works smoothly

Just driving by here... this looks like something that should be possible to add an automated test for.

SimonBarendse: Thanks @hculea, this indeed appears to be fixed now.

I like Marc's suggestion of adding a unit test for this. Could you do this?

hculea: Of course!

internals/secrethub/tree.go
SimonBarendse: The reason we've started with this change is to improve readability by making clear which "format" is used when it is printed.

The last change, switching to `format1`, `format2`, `format3` variables, goes back to having to lookup or guess what this format will be. Can we either use the format directly where it is printed (no variable) or change the name of the constants to something by which you can recognize it (I think this is really hard to get clear with naming, which is why I'm leaning towards defining the "format" strings directly in the print function call.

If you want to prevent duplication of the format strings between the secret and directory loop, can we extract that shared functionality between secrets and directories to a function to achieve that?

edif2008: Will push the version with the formats directly in the print commands first. Meanwhile, I will look up ways to extract the shared functionality

SimonBarendse: Sounds good :smile:

If extracting the shared functionality doesn't make the code more elegant, I'm also okay with having a small duplication.

edif2008: I looked into whether we can extract the shared functionality in an elegant way. Unfortunately, there are a couple of key differences that will not enable an elegant extraction: the difference between a secret and a directory, the fact that we add an extra `/` for the directory. This leaves us to only the checks which, although they look similar, if we extract them, we add a couple of more checks that will not make the code nice in the end (especially the readability). Therefore, we will leave it as it is, since the complexity is small and easy to understand and follow.

internals/secrethub/tree_test.go
```diff
@@ -0,0 +1,237 @@
...
+				path:      "test/repo",
+				fullPaths: true,
+				newClient: func() (secrethub.ClientInterface, error) {
+					return &secrethub.Client{}, nil
+				},
+			},
+			expectedOutput: "test/repo/\n" +
+				"├── test/repo/happy/\n" +
+				"├── test/repo/secretFolder/\n" +
+				"│   └── test/repo/secretFolder/found you\n" +
```
SimonBarendse, line 224: We'd expect this one to be red, right?

internals/secrethub/tree_test.go
```diff
@@ -0,0 +1,237 @@
...
+		"simple tree": {
+			cmd: &TreeCommand{
+				newClient: func() (secrethub.ClientInterface, error) {
+					return &secrethub.Client{}, nil
+				},
+			},
+			expectedOutput: "test/repo/\n" +
+				"├── happy/\n" +
+				"├── secretFolder/\n" +
+				"│   └── found you\n" +
```
SimonBarendse, line 209: We'd expect this one to be red, right?

internals/secrethub/tree_test.go
edif2008: Indeed. The tests have been adjusted to work properly

internals/secrethub/tree_test.go
edif2008: Indeed. The tests have been adjusted to work properly

internals/secrethub/tree.go
```diff
@@ -47,53 +50,90 @@ func (cmd *TreeCommand) Register(r command.Registerer) {
 	clause := r.Command("tree", "List contents of a directory in a tree-like format.")
 	clause.Arg("dir-path", "The path to to show contents for").Required().PlaceHolder(optionalDirPathPlaceHolder).SetValue(&cmd.path)
 
+	clause.Flag("full-paths", "Print the full paths of the directories and secrets.").Short('f').BoolVar(&cmd.fullPaths)
+	clause.Flag("no-indentation", "Print the content without indentation.").Short('i').BoolVar(&cmd.noIndentation)
+	clause.Flag("no-report", "Skip the report at the bottom.").BoolVar(&cmd.noReport)
+	clause.Flag("noreport", "Skip the report at the bottom.").BoolVar(&cmd.noReport)
```
SimonBarendse, line 56: What do you think about these changes to the help-texts for the flags?

I've taken the Unix `tree` command for inspiration.

```suggestion
	clause.Flag("full-paths", "Print the full path of each directory and secret.").Short('f').BoolVar(&cmd.fullPaths)
	clause.Flag("no-indentation", "Don't print indentation lines.").Short('i').BoolVar(&cmd.noIndentation)
	clause.Flag("no-report", "Turn off secret/directory count at end of tree listing.").BoolVar(&cmd.noReport)
	clause.Flag("noreport", "Turn off secret/directory count at end of tree listing.").BoolVar(&cmd.noReport)
```

SimonBarendse, line 56: While we do want to support both `--noreport` (as exists in the unix tree command) and `--no-report` (more consistent with how we name flags in other commands), we should only show one of them in the help-text. The one we're documenting is the "recommended way". The other we just add so that when you type that because you expect/remember it, it also works.

By excluding the other from the help-text, we keep the help-text concise and easier to understand.

```suggestion
	clause.Flag("no-report", "Skip the report at the bottom.").BoolVar(&cmd.noReport)
	clause.Flag("noreport", "Skip the report at the bottom.").Hidden().BoolVar(&cmd.noReport)
```

edif2008: The suggestion is great. Thanks

edif2008: Nice suggestion. Thanks

PR #322, opened 2020-07-24
Test suite for read command

edif2008:
This PR contains the necessary tests to ensure that the read command functions properly.

TO DO:
- [x] See how you can implement tests for clipboard on CIrcleCI

### DIFF
--- a/internals/secrethub/app.go
+++ b/internals/secrethub/app.go
@@ -2,6 +2,7 @@ package secrethub
 
 import (
 	"fmt"
+	"os"
 	"strings"
 	"text/template"
 
@@ -73,6 +74,9 @@ type App struct {
 
 // newClientFunc creates a ClientAdapater.
 type newClientFunc func() (secrethub.ClientInterface, error)
+
+// newWriterFunc creates a WriterAdapter
+type newWriterFunc func(filename string, data []byte, perm os.FileMode) error
 
 // NewApp creates a new command-line application.
 func NewApp() *App {
--- a/internals/secrethub/app.go
+++ b/internals/secrethub/app.go
@@ -2,7 +2,6 @@ package secrethub
 
 import (
 	"fmt"
-	"os"
 	"strings"
 	"text/template"
 
@@ -74,9 +73,6 @@ type App struct {
 
 // newClientFunc creates a ClientAdapater.
 type newClientFunc func() (secrethub.ClientInterface, error)
-
-// newWriterFunc creates a WriterAdapter
-type newWriterFunc func(filename string, data []byte, perm os.FileMode) error
 
 // NewApp creates a new command-line application.
 func NewApp() *App {
--- a/internals/secrethub/read.go
+++ b/internals/secrethub/read.go
@@ -28,7 +28,7 @@ type ReadCommand struct {
 	fileMode            filemode.FileMode
 	noNewLine           bool
 	newClient           newClientFunc
-	WriteFileFunc       func(filename string, data []byte, perm os.FileMode) error
+	writeFileFunc       func(filename string, data []byte, perm os.FileMode) error
 }
 
 // NewReadCommand creates a new ReadCommand.
@@ -38,7 +38,7 @@ func NewReadCommand(io ui.IO, newClient newClientFunc) *ReadCommand {
 		clearClipboardAfter: defaultClearClipboardAfter,
 		io:                  io,
 		newClient:           newClient,
-		WriteFileFunc:       ioutil.WriteFile,
+		writeFileFunc:       ioutil.WriteFile,
 	}
 }
 
@@ -92,7 +92,7 @@ func (cmd *ReadCommand) Run() error {
 	}
 
 	if cmd.outFile != "" {
-		err = cmd.WriteFileFunc(cmd.outFile, secretData, cmd.fileMode.FileMode())
+		err = cmd.writeFileFunc(cmd.outFile, secretData, cmd.fileMode.FileMode())
 		if err != nil {
 			return ErrCannotWrite(cmd.outFile, err)
 		}
--- a/internals/secrethub/read.go
+++ b/internals/secrethub/read.go
@@ -3,6 +3,7 @@ package secrethub
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
 	"time"
 
 	"github.com/secrethub/secrethub-cli/internals/cli/clip"
@@ -27,7 +28,7 @@ type ReadCommand struct {
 	fileMode            filemode.FileMode
 	noNewLine           bool
 	newClient           newClientFunc
-	newWriter           newWriterFunc
+	WriteFileFunc       func(filename string, data []byte, perm os.FileMode) error
 }
 
 // NewReadCommand creates a new ReadCommand.
@@ -37,7 +38,7 @@ func NewReadCommand(io ui.IO, newClient newClientFunc) *ReadCommand {
 		clearClipboardAfter: defaultClearClipboardAfter,
 		io:                  io,
 		newClient:           newClient,
-		newWriter:           ioutil.WriteFile,
+		WriteFileFunc:       ioutil.WriteFile,
 	}
 }
 
@@ -91,7 +92,7 @@ func (cmd *ReadCommand) Run() error {
 	}
 
 	if cmd.outFile != "" {
-		err = cmd.newWriter(cmd.outFile, secretData, cmd.fileMode.FileMode())
+		err = cmd.WriteFileFunc(cmd.outFile, secretData, cmd.fileMode.FileMode())
 		if err != nil {
 			return ErrCannotWrite(cmd.outFile, err)
 		}
--- a/internals/secrethub/read.go
+++ b/internals/secrethub/read.go
@@ -27,6 +27,7 @@ type ReadCommand struct {
 	fileMode            filemode.FileMode
 	noNewLine           bool
 	newClient           newClientFunc
+	newWriter           newWriterFunc
 }
 
 // NewReadCommand creates a new ReadCommand.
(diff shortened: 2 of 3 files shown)


internals/secrethub/read_test.go
```diff
@@ -1,9 +1,134 @@
...
+			},
+			err: testErr,
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			// Setup
+			io := fakeui.NewIO(t)
+			tc.cmd.io = io
```
mackenbach, line 105: Why make the separate `io` variable? Not only is it confusing (conflicts with the `io` package), it doesn't make sense to use it outside of the `tc.cmd` scope. Consider making it like this
```suggestion
			tc.cmd.io = fakeui.NewIO(t)
```

jpcoenen:
Task linked: [CU-6wtr69 Teest](https://app.clickup.com/t/6wtr69)

internals/secrethub/read_test.go
edif2008: The variable is made seperately because it is a mock and we can easily read the output that it has. I tried the suggestion you made, but I was no longer able to get the output's content. This happened due to the fact that the command accepts an interface, which means that I cannot use the `fakeui.FakeIO`'s specific fileds, in this case enabling me to see the output's content.
To solve this, I will rename the variable's name to avoid the conflict with the `io` package

internals/secrethub/read_test.go
```diff
@@ -1,9 +1,134 @@
...
+			cmd: ReadCommand{
+				path: "test/repo/secret",
+			},
+			versionService: fakeclient.SecretVersionService{
+				GetWithDataFunc: func(path string) (*api.SecretVersion, error) {
+					return &api.SecretVersion{
+						Data: []byte("testSecret"),
+					}, nil
+				},
+			},
```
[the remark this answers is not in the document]
edif2008: Nice catch. Tests have been improved

internals/secrethub/read_test.go
```diff
@@ -1,9 +1,134 @@
...
+				}
+			}
+
+			// Run
+			err := tc.cmd.Run()
+			if name == "success file" {
+				res, _ := ioutil.ReadFile(tc.cmd.outFile)
+				io.Out.WriteString(string(res))
+				os.Remove(tc.cmd.outFile)
+			}
```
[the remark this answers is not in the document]
edif2008: This is a very elaborate feedback. Really appreciate it.

1. Solved by providing a more general approach about the existence of the specific file.
2. Solved by the same approach as the one in 1
3. Solved by adding an extra feature which will contain the expected content of the clipboard/file
4. Renamed the variable to make it clearer and eliminate the conflict with the package name.

internals/secrethub/read_test.go
```diff
@@ -1,9 +1,122 @@
...
+				}
+			}
+
+			// Run
+			err := tc.cmd.Run()
+			content := ""
+			if _, err := os.Stat(tc.cmd.outFile); err == nil {
+				res, _ := ioutil.ReadFile(tc.cmd.outFile)
+				content = string(res)
+				os.Remove(tc.cmd.outFile)
```
Marton6, line 110: It is not recommended to use files in test cases for several reasons:

1. You might not clean up the file correctly if the test fails. In this case what would happen if `os.Stat` or `ioutil.ReadFile` returned an error? In either cases the out file wouldn't be deleted.

2. Your test cases are no longer isolated. Since all test cases would access the same file system, this would effectively behave like shared state and could bring about errors that are really hard to find and might even depend on the order of execution of the test cases.

3. It is generally good to have as much control over the outcome of the test as possible. In this case the calls to the os might sometimes fail for various reasons (no disk space left, a file with the same name already exists etc.). We only want the test to fail if our code is incorrect.

So to avoid this the best solution would be to replace the `ioutil.WriteFile` call in the read command's `Run` method with another function while testing, just like we're replacing `client.New` with other functions by setting the `newClientFunc` variable.

Note that in other situations, if writing to the file system is absolutely necessary, temporary files can be used when testing as these circumvent most of the issues listed here.

internals/secrethub/read_test.go
```diff
@@ -1,9 +1,122 @@
...
+			} else if tc.cmd.useClipboard {
+				res, _ := tc.cmd.clipper.ReadAll()
+				content = string(res)
+			}
```
Marton6, line 114: Instead of sometimes checking the clipboard and sometimes the output file, you could have expected values for all of these and check them all in every test case. (as we have recently discussed)

internals/secrethub/read_test.go
```diff
@@ -1,9 +1,122 @@
...
+			serviceErr: testErr,
+			err:        testErr,
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			// Setup
+			testIO := fakeui.NewIO(t)
+			tc.cmd.io = testIO
```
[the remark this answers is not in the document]
edif2008: Unfortunately, it will not. See https://github.com/secrethub/secrethub-cli/pull/322#discussion_r462725275 for the complete explanation.

internals/secrethub/read_test.go
```diff
@@ -1,9 +1,122 @@
...
+						SecretService: &fakeclient.SecretService{
+							VersionService: &fakeclient.SecretVersionService{
+								GetWithDataFunc: func(path string) (*api.SecretVersion, error) {
+									return &tc.secretVersion, tc.serviceErr
+								},
+							},
+						},
+					}, nil
+				}
+			}
```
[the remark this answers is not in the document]
edif2008: Just wanted to keep clear the two types of returns. However, it also works by having just one with your suggestion. Therefore, I will apply your idea. Thanks

internals/secrethub/read_test.go
edif2008: Done

internals/secrethub/read_test.go
mackenbach: 👍  on Marton's comment. Free tip: one way to use files in tests (when it is necessary) is by leveraging [ioutil.TempDir](https://golang.org/pkg/io/ioutil/#ReadDir)

internals/secrethub/read_test.go
```diff
@@ -1,9 +1,127 @@
...
+	testSecret := []byte("Remember! Reality's an illusion, the universe is a hologram, buy gold! Bye! - Bill Cipher")
+
+	cases := map[string]struct {
+		cmd             ReadCommand
+		newClientErr    error
+		secretVersion   api.SecretVersion
+		serviceErr      error
+		expectedClip    []byte
+		expectedFileOut []byte
+		out             string
```
mackenbach, line 30: For consistency, consider naming this field `expectedOut` or `expectedStdout`, whichever makes the most sense. Naming it `out` in this sense makes it look like a parameter, but it's actually part of the assertion. So matching the `actual` and `expected` pattern makes for more intuitive code.

internals/secrethub/read_test.go
edif2008: Done

internals/secrethub/read.go
```diff
@@ -36,6 +37,7 @@ func NewReadCommand(io ui.IO, newClient newClientFunc) *ReadCommand {
 		clearClipboardAfter: defaultClearClipboardAfter,
 		io:                  io,
 		newClient:           newClient,
+		newWriter:           ioutil.WriteFile,
```
Marton6, line 40: I think the name `newWriter` does not reflect what the function stored in this variable does.

If a function's name starts with `new` it implies that it is a constructor for something, in this case a writer, so I would expect it to return an `io.Writer` and take all parameters necessary to construct a writer.

In this case the function writes some data to a file, so I'd name it something like `writeFileFunc`. We have used the `Func`  suffix to denote that a variable is a function in situations like this.

internals/secrethub/app.go
```diff
@@ -74,6 +75,9 @@ type App struct {
 // newClientFunc creates a ClientAdapater.
 type newClientFunc func() (secrethub.ClientInterface, error)
 
+// newWriterFunc creates a WriterAdapter
+type newWriterFunc func(filename string, data []byte, perm os.FileMode) error
+
```
Marton6, line 80: I don't think it is necessary to declare a separate type for this function as it is only used in one place.

I'd say, for the sake of consistency we should drop this type.

internals/secrethub/read.go
edif2008: Fixed

internals/secrethub/app.go
edif2008: Fixed

internals/secrethub/read.go
```diff
@@ -27,6 +28,7 @@ type ReadCommand struct {
 	fileMode            filemode.FileMode
 	noNewLine           bool
 	newClient           newClientFunc
+	WriteFileFunc       func(filename string, data []byte, perm os.FileMode) error
```
Marton6, line 31: Is there a particular reason for exporting this field? I'd say we shouldn't export it unless we need to access it outside this package.

internals/secrethub/read_test.go
```diff
@@ -1,9 +1,130 @@
...
+	cases := map[string]struct {
+		cmd             ReadCommand
+		newClientErr    error
+		secretVersion   api.SecretVersion
+		serviceErr      error
+		expectedClip    []byte
+		expectedFileOut []byte
+		expectedOut     string
+		fileErr         error
+		err             error
```
Marton6, line 31: Consider naming this `expectedErr` if this is what we use in other tests.

internals/secrethub/read.go
edif2008: There is no reason. Probably by reflex of creating fields with capital letters. Will fix it

internals/secrethub/read_test.go
edif2008: in the other tests, `err` is the name of the expected error. I can change it to `expectedErr` for clarity (it may get confusing with `fileErr` as field), however, for consistency, I would leave it as it is. 
Let me know if you think I should change it despite the aspects mentioned above.

PR #324, opened 2020-07-27
Test suite for `credential disable` CLI command

hculea:


jpcoenen:
Task linked: [CU-6wtr69 Teest](https://app.clickup.com/t/6wtr69)

internals/secrethub/credential_disable_test.go
```diff
@@ -0,0 +1,141 @@
...
+			},
+			out: "A disabled credential can no longer be used to access SecretHub. " +
+				"This process can currently not be reversed.\nCredential disabled.\n",
+		},
+		"succeed-no-force": {
+			cmd: CredentialDisableCommand{
+				fingerprint: "62542d734d7f3627",
+				force:       false,
+			},
+			promptOut: "Are you sure you want to disable the credential with fingerprint 62542d734d7f3627? [y/N]: ",
```
Marton6, line 97: This fingerprint shows up in several places in this test function, therefore it would be a good idea to define it as a constant.

If it is used in a sentence you can use the [`fmt.Sprintf`](https://golang.org/pkg/fmt/#Sprintf) function to construct it.

internals/secrethub/credential_disable_test.go
```diff
@@ -0,0 +1,141 @@
...
+			in:        "y",
+			out:       "A disabled credential can no longer be used to access SecretHub. This process can currently not be reversed.\nCredential disabled.\n",
+		},
+		"fail-disable-error": {
+			cmd: CredentialDisableCommand{
+				fingerprint: "62542d734d7f3628",
+				force:       true,
+			},
+			out: "A disabled credential can no longer be used to access SecretHub. " +
+				"This process can currently not be reversed.\n",
```
Marton6, line 107: This long string could also be defined as a constant, as it has multiple occurrences.

internals/secrethub/credential_disable_test.go
hculea: This should now be solved! :D

internals/secrethub/credential_disable_test.go
hculea: This should now be solved! :D

internals/secrethub/credential_disable_test.go
```diff
@@ -0,0 +1,144 @@
...
+				fingerprint: validFingerprint,
+				force:       false,
+			},
+			promptOut: fmt.Sprintf("Are you sure you want to disable the credential with fingerprint %s? [y/N]: ", validFingerprint),
+			in:        "y",
+			out:       warningMessage + "Credential disabled.\n",
+		},
+		"fail disable error": {
+			cmd: CredentialDisableCommand{
+				fingerprint: "62542d734d7f3628",
```
Marton6, line 107: Why is this different from the `validFingerprint`?

go.mod
```diff
@@ -17,14 +17,12 @@ require (
 	github.com/op/go-logging v0.0.0-20160315200505-970db520ece7
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/secrethub/demo-app v0.1.0
-	github.com/secrethub/secrethub-go v0.30.0
+	github.com/secrethub/secrethub-go v0.29.1-0.20200728110331-9d7b31301226
```
Marton6, line 20: Why is the client version changed?

Marton6: Oh, got it. You use the fake credential service.

internals/secrethub/credential_disable_test.go
hculea: Because valid fingerprint is supposed to not return any error, while this one returns an error when the command tries to make a call to disable it.

hculea: What would be the second question? :)

edif2008: It is different because although it is a valid fingerprint (it can be used as a short version of a credential fingerprint), it is not the one that will be disabled since it doesn't exist.

Marton6: The other comment, but I figured it out after submitting the review :sweat_smile:

Marton6:
It seems that the client version should be changed to one which has the fake `CredentialIterator` implemented.

PR #327, opened 2020-07-28
Add getTree error case for 'acl ls' command

edif2008:
This PR contains an extra test case for the `acl ls` command: when getTree gets an error

### DIFF
--- a/internals/secrethub/acl_list_test.go
+++ b/internals/secrethub/acl_list_test.go
@@ -176,8 +176,8 @@ func TestACLListCommand_run(t *testing.T) {
 	for name, tc := range cases {
 		t.Run(name, func(t *testing.T) {
 			// Setup
-			io := fakeui.NewIO(t)
-			tc.cmd.io = io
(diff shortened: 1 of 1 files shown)


internals/secrethub/acl_list_test.go
```diff
@@ -176,8 +176,8 @@ func TestACLListCommand_run(t *testing.T) {
 	for name, tc := range cases {
 		t.Run(name, func(t *testing.T) {
 			// Setup
-			io := fakeui.NewIO(t)
-			tc.cmd.io = io
+			testIO := fakeui.NewIO(t)
+			tc.cmd.io = testIO
```
Marton6, line 180: Would this work?
```go
tc.cmd.io = fakeui.NewIO(t)
```

edif2008: Unfortunately, it will not. By doing what you suggested, I will no longer be able to get the output of the fake IO, since the `cmd.io` is an interface, while the fake one is a struct.

PR #328, opened 2020-07-29
Include test case for clipboard for the 'secrethub generate' command

edif2008:
This PR contains one extra test in which the clipboard is selected as the output

### DIFF
--- a/internals/secrethub/generate_test.go
+++ b/internals/secrethub/generate_test.go
@@ -21,6 +21,8 @@ func newIntValue(v int) intValue {
 
 func TestGenerateSecretCommand_run(t *testing.T) {
 	testErr := errio.Namespace("test").Code("test").Error("test error")
+	testData := []byte("random generated secret")
+	testPath := "namespace/repo/secret"
 
 	cases := map[string]struct {
 		cmd               GenerateSecretCommand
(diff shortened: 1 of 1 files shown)


internals/secrethub/generate_test.go
```diff
@@ -46,6 +47,24 @@ func TestGenerateSecretCommand_run(t *testing.T) {
...
+				copyToClipboard: true,
+				clipper:         fakeclip.New(),
+			},
+			writeFunc: func(path string, data []byte) (*api.SecretVersion, error) {
+				return &api.SecretVersion{Version: 1}, nil
+			},
+			path: "namespace/repo/secret",
+			data: []byte("random generated secret"),
+			out: "A randomly generated secret has been written to namespace/repo/secret:1.\n" +
+				"The generated value has been copied to the clipboard. It will be cleared after Less than a second.\n",
```
Marton6, line 66: Could we also check whether the value has actually been written to the fake clipboard?

edif2008: Done

internals/secrethub/generate_test.go
```diff
@@ -184,17 +221,16 @@ func TestGenerateSecretCommand_run(t *testing.T) {
 				}, tc.clientCreationErr
 			}
 
-			io := fakeui.NewIO(t)
-			tc.cmd.io = io
-
 			// Act
 			err := tc.cmd.run()
+			resClip, _ := tc.cmd.clipper.ReadAll()
```
Marton6, line 226: After this line you could call `assert.OK` on the error value, to make sure it is always nil.

edif2008: Good to know. 
Done

internals/secrethub/generate_test.go
```diff
@@ -20,111 +21,141 @@ func newIntValue(v int) intValue {
 
 func TestGenerateSecretCommand_run(t *testing.T) {
 	testErr := errio.Namespace("test").Code("test").Error("test error")
+	testData := []byte("random generated secret")
+	testPath := "namespace/repo/secret"
 
 	cases := map[string]struct {
 		cmd               GenerateSecretCommand
 		writeFunc         func(path string, data []byte) (*api.SecretVersion, error)
 		clientCreationErr error
```
Marton6, line 30: This should be named `newClientErr` so it is consistent with the naming in other tests.

edif2008: Fixed

PR #329, opened 2020-07-29
Testsuite for credential list

edif2008:
This PR contains the test suite for `credential ls` command

### DIFF
--- a/internals/secrethub/credential_list_test.go
+++ b/internals/secrethub/credential_list_test.go
@@ -4,31 +4,73 @@ import (
 	"testing"
 
 	"github.com/secrethub/secrethub-cli/internals/cli/ui/fakeui"
+	"github.com/secrethub/secrethub-go/internals/api"
 	"github.com/secrethub/secrethub-go/internals/assert"
 	"github.com/secrethub/secrethub-go/internals/errio"
 	"github.com/secrethub/secrethub-go/pkg/secrethub"
 	"github.com/secrethub/secrethub-go/pkg/secrethub/fakeclient"
+	"github.com/secrethub/secrethub-go/pkg/secrethub/iterator"
 )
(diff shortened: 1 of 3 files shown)


internals/secrethub/credential_list_test.go
```diff
@@ -0,0 +1,82 @@
...
+				tc.cmd.newClient = func() (secrethub.ClientInterface, error) {
+					return nil, tc.newClientErr
+				}
+			} else {
+				tc.cmd.newClient = func() (secrethub.ClientInterface, error) {
+					return fakeclient.Client{
+						CredentialService: &tc.credentialService,
+					}, nil
+				}
+			}
```
Marton6, line 72: Couldn't you just write:

```go
return fakeclient.Client{
    CredentialService: &tc.credentialService,
}, tc.newClientErr
```
?

internals/secrethub/credential_list_test.go
```diff
@@ -0,0 +1,82 @@
...
+	testErr := errio.Namespace("test").Code("test").Error("test error")
+
+	cases := map[string]struct {
+		cmd               CredentialListCommand
+		credentialService fakeclient.CredentialService
+		newClientErr      error
+		out               string
+		err               error
+	}{
+		"success list credentials": {
```
Marton6, line 24: Since we're testing the credential list command, this test case can be named just `success`

internals/secrethub/credential_list_test.go
edif2008: It works well. Thanks for the tip

internals/secrethub/credential_list_test.go
edif2008: Nice suggestion. Done

PR #330, opened 2020-07-29
Extended test suite for the `secrethub write` CLI command.

hculea:
Hereby, a test suite that achieves 80% in statement coverage, compared to the 65% achieved before.

### DIFF
--- a/internals/secrethub/write_test.go
+++ b/internals/secrethub/write_test.go
@@ -2,6 +2,7 @@ package secrethub
 
 import (
 	"bytes"
+	"errors"
 	"testing"
 
 	"github.com/secrethub/secrethub-cli/internals/cli/clip"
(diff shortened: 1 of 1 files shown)


edif2008:
Looks good for me as well

internals/secrethub/write_test.go
```diff
@@ -28,6 +29,7 @@ func TestWriteCommand_Run(t *testing.T) {
 		promptOut   string
 		promptErr   error
 		passwordIn  string
+		clientErr   error
```
Marton6, line 32: I suggest naming this error `newClientErr` because that's the name used in other tests.

internals/secrethub/write_test.go
```diff
@@ -192,6 +194,37 @@ func TestWriteCommand_Run(t *testing.T) {
 			},
 			err: errClipAndInFile,
 		},
+		"multiline and clip or in file": {
```
Marton6, line 197: It is not clear to me what `or in file` means in the name of this test.

If it is a test that sets both the multiline and clip flags, `multiline and clip` should suffice as a test name.

PR #331, opened 2020-08-01
Test critical CLI commands

edif2008:
It contains the following:

Tests created for the following commands:
 - `secrethub read`
 - `secrethub rm`
 - `secrethub service init --permission`
 - `secrethub credential ls`
 - `secrethub credential backup`

Additional tests made for the following commands:
 - `secrethub write`
 - `secrethub acl ls`
 - `secrethub acl check`
 - `secrethub generate`


internals/secrethub/credential_backup_test.go
```diff
@@ -0,0 +1,142 @@
...
+				return client, tc.newClientErr
+			}
+
+			err := tc.cmd.Run()
+
+			assert.Equal(t, err, tc.err)
+			// Since at the moment there is no way to retrieve the generated backup code
+			// we should just compare the beginning and the end of the message with the
+			// expected values, omitting, in this way, any assertion on the backup code.
+			if !tc.shouldSucceed {
```
jpcoenen, line 130: Would a sanity check on the code provide some value here? For example, checking if we get a 16 character string.

internals/secrethub/credential_list_test.go
```diff
@@ -0,0 +1,76 @@
...
+							},
+						},
+						CurrentIndex: 0,
+						Err:          nil,
+					}
+				},
+			},
+			out: "FINGERPRINT       TYPE  ENABLED  CREATED        DESCRIPTION\n" +
+				"8E146D837D4CA1DC  test  no       292 years ago  credential 1\n" +
+				"DFC3D1F0D9842F17        yes      292 years ago  credential 2\n",
```
jpcoenen, line 48: Will this test still succeed next year?

internals/secrethub/read_test.go
```diff
@@ -1,9 +1,130 @@
...
+	"github.com/secrethub/secrethub-go/internals/assert"
+	"github.com/secrethub/secrethub-go/internals/errio"
+	"github.com/secrethub/secrethub-go/pkg/secrethub"
+	"github.com/secrethub/secrethub-go/pkg/secrethub/fakeclient"
 )
 
 func TestReadCommand_Run(t *testing.T) {
-	// TODO SHDEV-1029 Test ReadCommand.
+	testErr := errio.Namespace("test").Code("test").Error("test error")
+	testSecret := []byte("Remember! Reality's an illusion, the universe is a hologram, buy gold! Bye! - Bill Cipher")
```
jpcoenen, line 20: Though I'm not into Gravity Falls (am I missing out on something?), gotta love the reference.

internals/secrethub/service_init_test.go
```diff
@@ -0,0 +1,244 @@
...
+				return fakeclient.Client{
+					ServiceService: &tc.serviceService,
+					AccessRuleService: &fakeclient.AccessRuleService{
+						SetFunc: func(path string, permission string, accountName string) (*api.AccessRule, error) {
+							returnedPerm, err := tc.setFunc(path, permission, accountName)
+							perm = returnedPerm
+							return returnedPerm, err
+						},
+					},
+				}, tc.err
```
jpcoenen, line 225: We're returning the exact same error here as we're checking on line 238, making the check on line 238 always succeed.

If we split these two in: `clientErr` and `expectedErr`, this becomes immediately clear.

internals/secrethub/service_init_test.go
```diff
@@ -0,0 +1,244 @@
...
+				credential: keyCreator,
+			},
+			serviceService: fakeclient.ServiceService{
+				CreateFunc: func(path string, description string, credentialCreator credentials.Creator) (*api.Service, error) {
+					return &api.Service{
+						ServiceID: "testService",
+					}, nil
+				},
+			},
+			err: ErrCannotWrite("path/test.txt", "open path/test.txt: no such file or directory"),
```
jpcoenen, line 76: We're not actually setting a `fileErr` in this test case. Should we?

jpcoenen (changes requested):
Very nice work guys! My apologies for the late review, the work deserves better :smile: 

Mostly looks good. Found a few things that could use some clarifications. 

My main overall suggestion is making really clear in the testcase setup what the inputs and the outputs are. That makes the cases a little easier to read and avoids problems where an expected output is used as an input (which makes the test case always succeed). See https://github.com/secrethub/secrethub-cli/pull/331#discussion_r490220492 and

internals/secrethub/read_test.go
edif2008: You got it right ^-^

internals/secrethub/credential_backup_test.go
edif2008: I added a check about the length of the credential. Besides that, I don't know what other sanity checks to add that may bring value.

internals/secrethub/credential_backup_test.go
```diff
@@ -0,0 +1,142 @@
...
+		promptOut     string
+		out           string
+		in            string
+		shouldSucceed bool
+		createFunc    func(creator credentials.Creator) *api.Credential
+		createErr     error
+		meErr         error
+		newClientErr  error
+		promptErr     error
+		err           error
```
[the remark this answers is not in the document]
edif2008: Made the distinction clearer.

internals/secrethub/credential_backup_test.go
```diff
@@ -0,0 +1,142 @@
...
+	var (
+		testErr    = errors.New("test")
+		failCreate = func(creator credentials.Creator) *api.Credential {
+			return &api.Credential{}
+		}
+		succeedCreate = func(creator credentials.Creator) *api.Credential {
+			_ = creator.Create()
+			return &api.Credential{}
+		}
+	)
```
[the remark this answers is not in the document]
edif2008: Indeed, it does not affect if we remove it. Also, it is clearer. 
@hculea, I will remove the `var` syntax. Let me know if there are reasons to put it back.

internals/secrethub/credential_list_test.go
edif2008: Good point. Adjusted so that they work anytime

internals/secrethub/service_init_test.go
edif2008: Indeed, there was a problem with the check. Adjusted the tests properly.

internals/secrethub/service_init_test.go
edif2008: Yes, we should. Fixed the test.

internals/secrethub/credential_backup_test.go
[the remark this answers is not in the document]
hculea: No, indeed, we can remove it without any issues. Thanks for handling it, Eddy.

jpcoenen (approved):
Yes, the changes look good!



For next time, I have one suggestion for committing: if you do a refactor like a rename (like with adding `expected[..]`), it makes reviewing really easy if that change is in a single commit. Then the reviewer can just eye-ball it to verify that you indeed ran the IDE's refactor command. If there are any other changes in the same commit, this makes this process a little more difficult. It was no big deal now (wasn't a huuge refactor fortunately :smile: ), but something to keep in mind next time :wink:

PR #332, opened 2020-08-07
secrethub init with setup code

Marton6:
This PR implements the `--setup-code` flag for the `secrethub init command` allowing users to initialize accounts using setup codes.


internals/secrethub/init.go
```diff
@@ -105,12 +121,98 @@ func (cmd *InitCommand) Run() error {
...
+			return err
+		}
+
+		err = cmd.credentialStore.ConfigDir().Credential().Write(encodedCredential)
+		if err != nil {
+			cmd.progressPrinter.Stop()
+			return err
+		}
+
+		client, err = cmd.newUnauthenticatedClient()
```
SimonBarendse, line 191: I can't come up with any way in which creating an unauthenticated client doesn't work, as it will try to load the credential from its default spots, but I believe it's safer to create a client already with the credential:

```suggestion
		client, err = cmd.newClientWithCredentials(credential)
```

It's certainly more explicit (and therefore easier to understand). And I believe it can also prevent an unneeded password prompt (and certainly prevents unneeded reading and decrypting).

internals/secrethub/client_factory.go
```diff
@@ -88,6 +89,18 @@ func (f *clientFactory) NewClientWithCredentials(provider credentials.Provider)
...
+	options := f.baseClientOptions()
+	options = append(options, secrethub.WithSetupCode(code))
+
+	client, err := secrethub.NewClient(options...)
+	if err != nil {
+		return nil, err
+	}
+
+	return client, nil
+}
```
jpcoenen, line 102: Shouldn't we just make `credentials.UseSetupCode` a credential? Then we do not have to add this function. Feels a bit superfluous because it is practically the same as above.

internals/secrethub/init.go
```diff
@@ -87,6 +100,7 @@ func (cmd *InitCommand) Run() error {
 		option, err := ui.Choose(cmd.io, "How do you want to initialize your SecretHub account on this device?",
 			[]string{
 				"Signup for a new account",
+				"Use a setup code to set up an account created through the website",
```
jpcoenen, line 103:
```suggestion
				"Use a setup code to set up an account created on the SecretHub website",
```

internals/secrethub/init.go
```diff
@@ -105,12 +121,98 @@ func (cmd *InitCommand) Run() error {
...
+		if err != nil {
+			cmd.progressPrinter.Stop()
+			return err
+		}
+
+		err = cmd.credentialStore.ConfigDir().Credential().Write(encodedCredential)
+		if err != nil {
+			cmd.progressPrinter.Stop()
+			return err
+		}
```
jpcoenen, line 189: Can we de-duplicate this code?

internals/secrethub/init.go
```diff
@@ -105,12 +121,98 @@ func (cmd *InitCommand) Run() error {
...
+		// Otherwise, we assume the passphrase was intentionally not
+		// configured to output a plaintext credential.
+		var passphrase string
+		if !cmd.credentialStore.IsPassphraseSet() && !cmd.force {
+			var err error
+			passphrase, err = ui.AskPassphrase(cmd.io, "Please enter a passphrase to protect your local credential (leave empty for no passphrase): ", "Enter the same passphrase again: ", 3)
+			if err != nil {
+				return err
+			}
+		}
```
jpcoenen, line 158: Can we de-duplicate this?

internals/secrethub/init.go
```diff
@@ -105,12 +121,98 @@ func (cmd *InitCommand) Run() error {
...
+
+		fmt.Fprintf(
+			cmd.io.Output(),
+			"An account credential will be generated and stored at %s. "+
+				"Losing this credential means you lose the ability to decrypt your secrets. "+
+				"So keep it safe.\n",
+			credentialPath,
+		)
```
jpcoenen, line 146: I think this is also in the signup, right? Can we de-duplicate it?

internals/secrethub/init.go
```diff
@@ -105,12 +121,98 @@ func (cmd *InitCommand) Run() error {
...
+		fmt.Fprint(cmd.io.Output(), "Setting up your account...")
+		cmd.progressPrinter.Start()
+
+		client, err := cmd.newClientWithSetupCode(setupCode)
+		if err != nil {
+			return err
+		}
+
+		credential := credentials.CreateKey()
+		_, err = client.Credentials().Create(credential, "")
```
jpcoenen, line 169: Would it be an idea to also ask for a device name when using a setup code? We did not do this before to not bother the user too much, but since username, email and full name are no longer asked in the CLI, I think we can add the question for device name (see the code for backup codes below). What do you think?

internals/secrethub/client_factory.go
Marton6: Simon and I thought that it makes more sense to not have it as a credential because it is only used for authentication, and not for decryption. Implying that it is a credential could be confusing for this reason. But I agree, in terms of code duplication it is not a good idea. What do you think?

jpcoenen: I think that if we make sure that the client can handle this situation correctly (so no nil dereference panics) and it is documented, there should be no real problem. Using anything from the `credentials` package requires some documentation anyway. For example, you cannot use `credentials.UseAWS()` without knowing that this requires an AWS role to be assumed.

internals/secrethub/init.go
Marton6: I couldn't come up with a meaningful method for deduplicating this. Maybe we could factor out all the prompting and credential writing the same way we moved the environment related stuff out of the run command.

internals/secrethub/init.go
Marton6: I made the message a constant, which is the most meaningful way I could deduplicate this.

internals/secrethub/init.go
Marton6: Good idea. I added the prompt (and deduplicated it)

internals/secrethub/init.go
jpcoenen: Hmm, okay, de-duplicating the control logic is maybe not straightforward nor would it improve the code quality. But then we should at least de-duplicate the strings used. Either by making them constants or - I think preferably - moving this to a helper function that is called here and in the backup code initialization. Something like:
```golang
func askCredentialPassphrase(io ui.IO) (error, string){
    return ui.AskPassphrase(io, "Please enter a passphrase to protect your local credential (leave empty for no passphrase): ", "Enter the same passphrase again: ", 3)
}
```

internals/secrethub/init.go
```diff
@@ -162,27 +162,32 @@ func (cmd *InitCommand) Run() error {
 
 		client, err := cmd.newClientWithSetupCode(setupCode)
 		if err != nil {
+			cmd.progressPrinter.Stop()
```
jpcoenen, line 164: Not something to fix now, but I think we would make our lives a lot easier if we'd just allow `progressPrinter.Stop()` to be called multiple times. Then we can just defer the stopping.

internals/secrethub/init.go
Marton6: Good idea! I implemented it.

internals/secrethub/init.go
```diff
@@ -105,12 +119,90 @@ func (cmd *InitCommand) Run() error {
...
+
+		fmt.Fprintf(cmd.io.Output(), credentialCreationMessage, credentialPath)
+
+		// Only prompt for a passphrase when the user hasn't used --force.
+		// Otherwise, we assume the passphrase was intentionally not
+		// configured to output a plaintext credential.
+		var passphrase string
+		if !cmd.credentialStore.IsPassphraseSet() && !cmd.force {
+			var err error
+			passphrase, err = ui.AskPassphrase(cmd.io, "Please enter a passphrase to protect your local credential (leave empty for no passphrase): ", "Enter the same passphrase again: ", 3)
```
florisvdg, line 146: Same thing here too?
```suggestion
			passphrase, err = askCredentialPassphrase(cmd.io)
```

jpcoenen (approved):

PR #340, opened 2020-09-04
Refer to web signup in README

jpcoenen:


README.md
```diff
@@ -23,7 +23,7 @@ The SecretHub CLI provides the command-line interface to interact with the Secre
 
 ## Get started
 
-Follow the [Getting Started Guide][getting-started] to quickly get up and running with SecretHub :rocket:
+[Set up SecretHub on your machine](https://signup.secrethub.io/) to quickly get up and running :rocket:
```
florisvdg, line 26: I think this entire section can go, because we have the `Get started` button right above it. 

If we think that one button isn't enough, we can even add that button again after the `Usage` section.

florisvdg: Tried it out here b9444ac and looks good to me.

florisvdg (approved):

PR #341, opened 2020-09-07
Remove prompt for setup code

jpcoenen:
There is no valid use case where a user would have to enter it manually,
so we can better remove this altogehter and direct users to the website.

### DIFF
--- a/internals/secrethub/init.go
+++ b/internals/secrethub/init.go
@@ -6,8 +6,6 @@ import (
 	"os"
 	"time"
 
-	"github.com/secrethub/secrethub-go/internals/api"
-
 	"github.com/secrethub/secrethub-cli/internals/cli/progress"
 	"github.com/secrethub/secrethub-cli/internals/cli/ui"
 	"github.com/secrethub/secrethub-cli/internals/secrethub/command"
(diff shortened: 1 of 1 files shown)


internals/secrethub/init.go
```diff
@@ -128,11 +126,8 @@ func (cmd *InitCommand) Run() error {
 	case InitModeSetupCode:
 		setupCode := cmd.setupCode
 		if setupCode == "" {
-			var err error
-			setupCode, err = ui.AskAndValidate(cmd.io, "What is your setup code?\n", 3, api.ValidateSetupCode)
-			if err != nil {
-				return err
-			}
```
florisvdg, line 135: Now only the additional prompt is gone, but we can also delist/hide setup codes entirely from the `secrethub init` options, so remove this too: https://github.com/secrethub/secrethub-cli/blob/465b8da80e898acfef90fb1653f4c2571204d354/internals/secrethub/init.go#L99

jpcoenen: Sure. That works together nicely with https://github.com/secrethub/secrethub-cli/pull/342

florisvdg (approved):
👍

PR #344, opened 2020-09-07
Fix incorrect int to string conversion 

jpcoenen:
This line breaks compilation under go 1.15

### DIFF
--- a/internals/cli/filemode/filemode.go
+++ b/internals/cli/filemode/filemode.go
@@ -2,7 +2,6 @@
 package filemode
 
 import (
-	"fmt"
 	"os"
 	"strconv"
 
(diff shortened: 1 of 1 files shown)


SimonBarendse (commented):
@jpcoenen I've fixed one issue (explained in the commit message). If you agree you can merge this.

jpcoenen:
Apparently it was not as straightforward as I thought.

SimonBarendse:
> Previous implementation outputted something like -rw-r--r-- instead of 0644.

Isn't that a feature?

SimonBarendse (commented):
> Previous implementation outputted something like -rw-r--r-- instead of 0644.

Isn't that a feature?

jpcoenen:
> Isn't that a feature?

Shouldn't it be the inverse of the `Set()` function?

SimonBarendse:
Hmm, I'm not sure whether that's required.

I'm checking the docs now at https://golang.org/pkg/flag/#Value , but don't see anything mentioned about that.

jpcoenen:
Still, why would we? We use the digit-format everywhere.

SimonBarendse (approved):
Yes, that's right. Doesn't hurt to have it like this.

PR #357, opened 2020-12-02
Move update-passphrase command to "credential" management command

SimonBarendse:
Now that we have a `credential` management command, I believe it's more intuitive to group `update-passphrase` in there as well. What do you think?

This deprecates `secrethub config update-passphrase` command, in favor of `secrethub credential update-passphrase`.


jpcoenen (approved):
Way better :+1:

PR #358, opened 2020-12-04
Improve credential errors

Marton6:
This PR, in combination with https://github.com/secrethub/secrethub-go/pull/223 ensures that all credential related errors are wrapped with a message that includes the source of the credential (except if the credential is passed via a cli flag).

### DIFF
--- a/internals/secrethub/credential_store.go
+++ b/internals/secrethub/credential_store.go
@@ -51,6 +51,7 @@ func (store *credentialConfig) IsPassphraseSet() bool {
 }
 
 // Register registers the flags for configuring the store on the provided Registerer.
+// The environment variables of these flags are also checked on the client, but checking them here allows us to fail fast.
 func (store *credentialConfig) Register(r FlagRegisterer) {
 	r.Flag("config-dir", "The absolute path to a custom configuration directory. Defaults to $HOME/.secrethub").Default("").PlaceHolder("CONFIG-DIR").SetValue(&store.configDir)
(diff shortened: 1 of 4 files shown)


internals/secrethub/credential_store.go
```diff
@@ -69,10 +74,13 @@ func (store *credentialConfig) Import() (credentials.Key, error) {
 }
 
 func (store *credentialConfig) getCredentialReader() credentials.Reader {
-	if store.AccountCredential != "" {
-		return credentials.FromString(store.AccountCredential)
+	if store.AccountCredential == "" {
+		return store.configDir.Credential()
+	}
+	if store.credentialFlag.HasEnvarValue() {
+		return credentials.FromEnv("SECRETHUB_CREDENTIAL")
```
SimonBarendse, line 81: It feels a bit awkward to have the flag already load the credential from the environment variable and validate it, but then load the credential from envvar again here and not use the already loaded value in `store.AccountCredential`.

We're doing all this to pass the source of the credential, did I get that right? If so, I'm wondering whether we can have the flag itself implement `CredentialSource` interface introduced in the accompanying PR https://github.com/secrethub/secrethub-go/pull/223 and return the flag itself as the credential reader?

The `Source` function of the credential flag could use `HasEnvarValue` to determine what to return.

Marton6: Great idea! I implemented an (manually) tested it and it works well. The code is also much cleaner, thanks :+1:

Marton6:
This PR now depends on https://github.com/secrethub/secrethub-go/pull/223 being merged first.

Marton6:
PR was closed by mistake.

SimonBarendse (approved):

PR #359, opened 2020-12-16
account inspect for service accounts

Marton6:
This PR makes the `secrethub account inspect` command work for service accounts. The output of the command is extended to allow for easier parsing. The following fields are outputted:

for services:
- CreatedAt
- Description
- AccountType
- AccountName
- PublicAccountKey

for users:
- Username
- FullName
- Email
- EmailVerified
- CreatedAt
- AccountType
- AccountName
- PublicAccountKey

### DIFF
--- a/internals/secrethub/account_inspect.go
+++ b/internals/secrethub/account_inspect.go
@@ -107,7 +107,6 @@ func newOutputUser(user *api.User, timeFormatter TimeFormatter) *outputUser {
 // outputService is a user friendly JSON representation of a service account.
 type outputService struct {
 	Description string
-	CreatedAt   string `json:",omitempty"`
 	outputAccount
 }
 
--- a/internals/secrethub/account_inspect.go
+++ b/internals/secrethub/account_inspect.go
@@ -40,38 +40,84 @@ func (cmd *AccountInspectCommand) Run() error {
 		return err
 	}
 
-	user, err := client.Users().Me()
+	account, err := client.Accounts().Me()
 	if err != nil {
 		return err
 	}
-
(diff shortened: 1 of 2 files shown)


internals/secrethub/account_inspect.go
```diff
@@ -40,38 +43,84 @@ func (cmd *AccountInspectCommand) Run() error {
...
+	Description string
+	CreatedBy   string
+	CreatedAt   string `json:",omitempty"`
+	outputAccount
+}
+
+func newOutputService(service *api.Service, account *api.Account, timeFormatter TimeFormatter) *outputService {
+	return &outputService{
+		Description: service.Description,
+		CreatedBy:   service.CreatedBy.String(),
```
Marton6, line 118: This will be a `uuid` is this an issue?

jpcoenen: I don't think an UUID has any use for a user. So what can we do to solve that?

Marton6: Thought so too. Having the username there would be way more useful, but to do that we'd have to either:
1. Modify the server controller for this request to (also) send the username.
2. Add another route to the server that gets a user by its id and call that (which sounds like a pretty bad idea to me)

What do you think?

jpcoenen: Another option I just thought of: we can also leave it out altogether (at least for now). I think the use-case for getting this info about the service account you're currently using, is pretty limited.

Marton6: Good point. I agree, let's leave it out for now.

jpcoenen (approved):

PR #361, opened 2020-12-16
Add secrethub binary to the gitignore

SimonBarendse:
To prevent it from being committed as in this commit ec7fd122919b4bcd5159681e9b173123407e2802

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
-secrethub
+# Ignore binary
+/secrethub
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+secrethub


.gitignore
```diff
@@ -0,0 +1 @@
+secrethub
```
Marton6, line 1: I think this will match (recursively) all files/dirs named 'secrethub' not just the one on the `.gitignore` level. We should add a `/` to the beginning of the line to only match files named secrethub that are next to the `.gitignore`.

> If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself. Otherwise the pattern may also match at any level below the .gitignore level.

Source: https://git-scm.com/docs/gitignore

SimonBarendse: Good spot!

Marton6 (approved):

PR #362, opened 2020-12-23
Allow configuring flags from env vars

Marton6:
This PR adds a pre-run hook to all cobra commands that sets all unset flags to their corresponding environment variable's value. This also works in the case of persistent flags.

This PR introduces the limitation of registering all persistent flags on the root command.

### DIFF
--- a/internals/cli/env.go
+++ b/internals/cli/env.go
@@ -76,6 +76,12 @@ func (a *App) Command(name, help string) *CommandClause {
 	return clause
 }
 
+// PersistentFlags returns a flag set that allows configuring
+// global persistent flags (that work on all commands of the CLI).
+func (a *App) PersistentFlags() *FlagSet {
+	return &FlagSet{FlagSet: a.Cmd.Cmd.PersistentFlags(), cmd: a.Cmd}
+}
+
 // Version adds a flag for displaying the application version number.
 func (a *App) Version(version string) *App {
 	a.Cmd.Cmd.Version = version
@@ -290,10 +296,6 @@ func (c *CommandClause) Flag(name string) *Flag {
 
 func (c *CommandClause) Flags() *FlagSet {
 	return &FlagSet{FlagSet: c.Cmd.Flags(), cmd: c}
-}
-
-func (a *App) PersistentFlags() *FlagSet {
-	return &FlagSet{FlagSet: a.Cmd.Cmd.PersistentFlags(), cmd: a.Cmd}
 }
 
 // BindArguments binds a function to a command clause, so that
--- a/internals/cli/env.go
+++ b/internals/cli/env.go
@@ -37,7 +37,7 @@ type App struct {
 func NewApp(name, help string) *App {
 	app := &App{
 		Cmd: &CommandClause{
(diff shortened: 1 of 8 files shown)


internals/secrethub/debug.go
```diff
@@ -8,15 +8,10 @@ import (
...
-	commandClause := cli.CommandClause{
-		Cmd: app.Cmd,
-		App: app,
-	}
 	flag := debugFlag{
 		logger: logger,
 	}
-	commandClause.PersistentFlags().VarP(&flag, "debug", "D", "Enable debug mode.")
-	commandClause.Flag("debug").NoOptDefVal = "true"
+	app.PersistentFlags().VarP(&flag, "debug", "D", "Enable debug mode.")
```
edif2008, line 14: This flag no longer works without an argument, which means that it cannot act like a bool flag.
```suggestion
	app.PersistentFlags().VarP(&flag, "debug", "D", "Enable debug mode.")
	app.Cmd.Flag("debug").NoOptDefVal = "true"
```

internals/secrethub/mlock.go
```diff
@@ -29,13 +29,8 @@ func (f mlockFlag) init() error {
...
 // RegisterMlockFlag registers a mlock flag that enables memory locking when set to true.
 func RegisterMlockFlag(app *cli.App) {
-	commandClause := cli.CommandClause{
-		Cmd: app.Cmd,
-		App: app,
-	}
 	flag := mlockFlag(false)
-	commandClause.PersistentFlags().Var(&flag, "mlock", "Enable memory locking")
-	commandClause.Flag("mlock").NoOptDefVal = "true"
+	app.PersistentFlags().Var(&flag, "mlock", "Enable memory locking")
```
edif2008, line 33: This flag no longer works without an argument, which means that it cannot act like a bool flag.
```suggestion
	app.PersistentFlags().Var(&flag, "mlock", "Enable memory locking")
	app.Cmd.Flag("mlock").NoOptDefVal = "true"
```

internals/cli/env.go
```diff
@@ -25,7 +25,7 @@ var (
 // App represents a command-line application that wraps the
 // kingpin library and adds additional functionality.
 type App struct {
-	Cmd              *cobra.Command
+	Cmd              *CommandClause
```
edif2008, line 28: May be a bit weird to suggest, but since it is the root commands, should we rename it to `Root`?
I ask this because I noticed some code with `app.Cmd.Cmd`, which can be confusing. Doing `app.Root.Cmd` may look a bit cleaner. Let me know what you think.

edif2008 (commented):
These adjustments are just making the quality of the cli way better. Great job. ^-^ 
Take a look at the comments and once solved, we can merge this.

internals/cli/env.go
Marton6: Yeah, good point. I think `Root` sounds good.

Marton6: I'll probably simplify this in a refactor PR

edif2008 (approved):

PR #363, opened 2020-12-23
Refactor cobra code

Marton6:
This PR splits up `env.go` into `app.go`, `argument.go` and `flag.go` so the code can be managed more easily.

A couple typos from comments have also been fixed.


internals/cli/argument.go
```diff
@@ -0,0 +1,103 @@
...
+func shortDur(d *time.Duration) string {
+	s := d.String()
+	if strings.HasSuffix(s, "m0s") {
+		s = s[:len(s)-2]
+	}
+	if strings.HasSuffix(s, "h0m") {
+		s = s[:len(s)-2]
+	}
+	return s
+}
```
edif2008, line 103: This is used by `DurationVarP` and `DurationVar` from `flag.go`. Therefore, it should be moved there.

edif2008 (commented):
Nice division of the 4 custom objects that we made.
One small adjustment, and this can be merged.

internals/cli/argument.go
Marton6: Oh, that's right

Marton6: Fixed

edif2008 (approved):

PR #364, opened 2020-12-24
Install autocompletion for cobra-cli

edif2008:
This enables the user to install autocompletion using `secrethub completion <shell>`
Note that for `zsh` and `bash`, sudo privileges are required. We let the user be aware of this before performing installation.

Installation is provided for `bash` (both linux and MacOS), `zsh`, `fish` and `powershell`.


mackenbach:
Hi @edif2008 this looks like a lot of commits for a simple autocompletion install command. Should you be merging this into a feature branch instead or do a rebase or something?

edif2008:
My bad. It was supposed to be for `cobra-switch` branch. Thank you for noticing

internals/secrethub/completion.go
```diff
@@ -21,112 +32,174 @@ func NewCompletionCommand() *CompletionCommand {
 
 // Register registers the command, arguments and flags on the provided Registerer.
 func (cmd *CompletionCommand) Register(r cli.Registerer) {
-	cmd.clause = r.Command("autocomplete", "Generate completion script")
+	cmd.clause = r.Command("completion", "Installs completion script").Hidden()
 	cmd.clause.Cmd.DisableFlagsInUseLine = true
+	cmd.clause.Flags().BoolP("n", "n", false, "generate script now")
```
jpcoenen, line 37: What does this do?

jpcoenen (commented):
First of all: very nice that you got it working for all shells and OS's. 

I have one general remark about the current approach: we're currently using (power)shell scripts and commands within Go. This means we're creating a very tight coupling with whatever is available within the OS, without allowing the user to modify any of it. 

I'd suggest we go for one of these options:
1. Either use native Go (so writing files, reading environment variables etc with the `os` package).
2. Just provide the shell scripts to the user in our documentation. That way, they have full control over what is happening and are able to make any changes if necessary. It also gives more control over when `sudo` is called.

Since we already have all the scripts needed, I'd say option 2 is by far the easiest. Then in the next release we can ship an auto-install option written in native Go.

edif2008:
Based on the discussion with @jpcoenen, this solution will not work for the long run, since every time a new command or flag is implemented, the user will have to redo this command to update the completion script installed on their device. 

To overcome this, we will need to provide completion in the install package, which needs more time. A similar approach is made by both `docker` and `helm`.

PR #369, opened 2021-02-05
Remove unauthenticated client

jpcoenen:
This is no longer necessary now that the signup command is removed.


SimonBarendse (approved):
Nice cleanup! :ok_hand:

PR #370, opened 2021-02-11
Improve credential not found error for service accounts

SimonBarendse:
The previous error message was solely focused on setting up a credential for a human. This new error message attempts to cover  both the human and service account use-case.


jpcoenen (approved):

SimonBarendse:
I've opened #371 to fix the lint job.

PR #375, opened 2021-03-31
Account delete

Marton6:
This PR adds the `secrethub account delete` command, which deletes the authenticated account if it is not a member of any orgs.

This branch uses the version of the go client from this PR: https://github.com/secrethub/secrethub-go/pull/237/files


internals/secrethub/account_delete.go
```diff
@@ -0,0 +1,91 @@
...
+		fmt.Fprintln(cmd.io.Output(), "Name does not match. Aborting.")
+		return nil
+	}
+
+	err = client.Accounts().Delete(account.AccountID)
+	if err != nil {
+		return err
+	}
+	fmt.Fprintln(cmd.io.Output(), "Your account was successfully deleted.")
+	return nil
```
Marton6, line 90: We might also want to remove the credential file after account deletion, since, otherwise you get an invalid credential error whenever you try to run any other command that requires the credential.

SimonBarendse: Yes, let's do that :+1:

internals/secrethub/account_delete.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	if len(orgList) > 0 {
+		builder := strings.Builder{}
+		for i, org := range orgList {
+			builder.WriteString(org.Name)
+			if i != len(orgList)-1 {
+				builder.WriteString(", ")
+			}
+		}
+		return fmt.Errorf("cannot delete account that is a member of an organization. Please leave or delete the following organizations before deleting your account: %s", builder.String())
```
SimonBarendse, line 66: To make this more user-friendly we should do one of two things:
1. Make it possible to interactively leave or delete orgs from the `account delete` command.
2. Provide clearer instructions with the exact commands to execute to leave or delete orgs.

internals/secrethub/account_delete.go
```diff
@@ -0,0 +1,91 @@
...
+func NewAccountDeleteCommand(io ui.IO, newClient newClientFunc) *AccountDeleteCommand {
+	return &AccountDeleteCommand{
+		io:        io,
+		newClient: newClient,
+	}
+}
+
+// Register registers the command, arguments and flags on the provided Registerer.
+func (cmd *AccountDeleteCommand) Register(r cli.Registerer) {
+	clause := r.Command("delete", "Delete your SecretHub account.")
```
SimonBarendse, line 29: We could consider adding a `HelpLong` with information on the prerequisites (e.g. leave/delete orgs) and the impact of the command (permanent deletion of all secrets/repos/dirs etc. without any means of restoring)

SimonBarendse (commented):
Very nice to get this self-serve :+1: This will help users who have finished their migration to 1Password and want to delete their data.

internals/secrethub/account_delete.go
Marton6: Shall I implement an interactive "wizard" that handles this? Something like:
```
if user is not last org admin {
    leave org
} else if user is last member of org {
    delete org
} else  {
    choose someone else to be admin instead of you
}
```

This would be done for each of the orgs that you are a member of, and, of course it would prompt for confirmation (type the org name to confirm, etc.).

Does this look good to you?

SimonBarendse: Yep, that looks about right :+1:

I'm wondering if removing the org should be an option in all scenarios:
```
if user is last member of org {
    prompt with confirmation to remove the org
} else if other admin exists {
    prompt with question to leave or remove the org
} else {
    prompt with question to promote someone to admin or remove the org
}
```

Since depending on what scenario you're in for removing your account, there may not be a use case for handing over the org anymore. For example: if you're an admin of an org that you've migrated to 1Password, you probably don't want to promote someone else to the admin role, but just want to remove the data as now 1Password is the source of truth for it.

What do you think?

Marton6: Yupp :+1:  I had the same idea but couldn't come up with a use case for it. Migrating to 1Password sounds like a good use case to support with this prompt.

Marton6: I slightly deviated from this plan but I think the result makes more sense.

SimonBarendse (approved):
Awesome work!

internals/secrethub/account_delete.go
```diff
@@ -86,6 +89,12 @@ func (cmd *AccountDeleteCommand) Run() error {
 	if err != nil {
 		return err
 	}
+	if cmd.credentialStore.Source() == CredentialSourceConfigDir {
+		err := os.Remove(cmd.credentialStore.ConfigDir().Credential().Path())
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "warning: could not delete credential file: %v", err)
```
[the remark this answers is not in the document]
Marton6: Good idea, I updated the message

internals/secrethub/account_delete.go
Marton6: Added a help long.

PR #380, opened 2021-04-23
Skip path if the person in org does not have any permissions

edif2008:
Before the migration tool would fail if the path was not valid. Sometimes, the path is valid (e.g. a repo), however the user does not have any permissions set for that specific repo, even though they are a member of the organization owning the repo.

Now the migration tool will just skip it with a warning.

### DIFF
--- a/internals/secrethub/migrate.go
+++ b/internals/secrethub/migrate.go
@@ -272,6 +272,9 @@ func (cmd *MigratePlanCommand) addDirToPlan(client secrethub.ClientInterface, pa
 		}
 		return nil
 	}
+	if err != nil {
+		return err
+	}
 
 	err = addTreeToPlan(tree, plan)
 	if err != nil {
--- a/internals/secrethub/migrate.go
+++ b/internals/secrethub/migrate.go
@@ -256,7 +256,7 @@ func (cmd *MigratePlanCommand) addDirToPlan(client secrethub.ClientInterface, pa
 	fmt.Fprintf(cmd.io.Output(), "Planning migration for %s\n", path)
 
 	tree, err := client.Dirs().GetTree(path, -1, false)
-	if err == api.ErrForbidden {
+	if err == api.ErrForbidden || api.IsErrNotFound(err) {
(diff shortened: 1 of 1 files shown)


internals/secrethub/migrate.go
```diff
@@ -257,7 +257,7 @@ func (cmd *MigratePlanCommand) addDirToPlan(client secrethub.ClientInterface, pa
 
 	tree, err := client.Dirs().GetTree(path, -1, false)
 	if err == api.ErrForbidden {
```
SimonBarendse, line 259: Could we make this `if err == api.ErrForbidden || api.IsErrNotFound(err)`, so that we don't include any other errors than those caused by missing read permission? Then we don't require to say `if ... then ....` in the error message.

SimonBarendse (commented):
Thanks for opening this! :+1:

Marton6 (approved):
LGTM. It fixes the issue that I encountered when running the migration tool on an org with a repo that I didn't have access to.

internals/secrethub/migrate.go
edif2008: Nice catch. Adjusted properly.

internals/secrethub/migrate.go
```diff
@@ -272,9 +272,6 @@ func (cmd *MigratePlanCommand) addDirToPlan(client secrethub.ClientInterface, pa
 		}
 		return nil
 	}
-	if err != nil {
```
SimonBarendse, line 275: This err is now unhandled.

We should still keep

```go
if err != nil {
    return err
}
```

for the cases where the error is not forbidden or not found.

SimonBarendse (changes requested):

internals/secrethub/migrate.go
edif2008: Adjusted properly.

Marton6 (approved):

SimonBarendse (approved):
:rocket:

Thank you!

PR #384, opened 2021-05-07
Add integration migration instructions to the apply success message

SimonBarendse:
When you're done applying the secrets migration, a logical next step would be to migrate your integrations. Including the link in the  success message makes it easier to follow along with the migration process.


SimonBarendse:
Thanks for adding that fix! :heart:

PR #386, opened 2021-07-13
Create config migration commands for run and inject

florisvdg:
### Migrate config template files
```
secrethub migrate config templates <path/to/tpl>
```

### Migrate `secrethub.env` files
```
secrethub migrate config envfile <path/to/secrethub.env>
```

### Migrate `secrethub://` references in configuration code
```
secrethub migrate config references <path/to/file>
```

### Examples with template variables
```
secrethub migrate config templates --var env=dev,staging,prod --var region=us-east-1,eu-west-1 my-config.yml.tpl
```

```
secrethub migrate config envfile --var env=dev,staging,prod --var region=us-east-1,eu-west-1 secrethub.env
```

Note: all config migration commands should be considered 'best effort'.

PR #388, opened 2021-08-23
Release v0.43.0

SimonBarendse:
### Added
- Added commands that allows you to migrate secrethub.env files, configuration files that include secret references and template files (used with `inject`) to 1Password. See https://secrethub.io/docs/1password/migration/ for detailed instructions on how to use these new commands. (#386)

### Changed
- The command to migrate your secrets from SecretHub to 1Password is now idempotent, so it supports running them again to apply any new changes without changing the already migrated secrets. (#379)
- When migrating your secrets to 1Password, the API Credential item type is now used instead of the Login item type (#387)


artbenito (commented):
****

PR #389, opened 2021-08-23
Fix `migrate config templates` command

edif2008:
The command would throw an error because it did not have write permission for overwriting the provided file.

### DIFF
--- a/internals/secrethub/migrate_config_templates.go
+++ b/internals/secrethub/migrate_config_templates.go
@@ -1,10 +1,10 @@
 package secrethub
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"io/ioutil"
-	"os"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


SimonBarendse (approved):

internals/secrethub/migrate_config_templates.go
```diff
@@ -48,7 +47,7 @@ func (cmd *MigrateConfigTemplatesCommand) Run() error {
 	return nil
 }
 
-func migrateTemplateTags(inFile io.Reader, outFile io.Writer, mapping referenceMapping, formatString string) ([]string, error) {
+func migrateTemplateTags(inFile io.Reader, outFile string, mapping referenceMapping, formatString string) ([]string, error) {
```
florisvdg, line 50: I'd say let's take the same approach here with the signature as the `migrateReferences` function and just work with the file contents string, because mixing filepaths with io readers is a bit confusing.

internals/secrethub/migrate_config_templates.go
```diff
@@ -89,7 +88,7 @@ func migrateTemplateTags(inFile io.Reader, outFile io.Writer, mapping referenceM
 		return nil, fmt.Errorf(errMsg)
 	}
 
-	_, err = io.WriteString(outFile, output)
+	err = ioutil.WriteFile(outFile, []byte(output), 0666)
```
florisvdg, line 91: Ideally, it would preserve the file mode of the existing file.

edif2008: Adjusted to preserve file mode

internals/secrethub/migrate_config_templates.go
edif2008: Adjusted the signature accordingly.

SimonBarendse (approved):

florisvdg (approved):

PR #390, opened 2021-08-24
Enable migrating templates without whitespaces

edif2008:
Now it is no longer enforced to have template formats with whitespaces between the path and the curly brackets, such as 
`{{ path/to/secrets }}`

Depends on #389

### DIFF
--- a/internals/secrethub/migrate_config_envfile.go
+++ b/internals/secrethub/migrate_config_envfile.go
@@ -39,7 +39,7 @@ func (cmd *MigrateConfigEnvfileCommand) Run() error {
 		return err
 	}
 
-	replaceCount, err := migrateTemplateTags(bytes.NewBuffer(inFileContents), ".env", refMapping, "%s")
+	replaceCount, err := migrateTemplateTags(bytes.NewBuffer(inFileContents), ".env", refMapping, "{{ %s }}")
 	if err != nil {
 		return err
 	}
--- a/internals/secrethub/migrate_config_envfile.go
+++ b/internals/secrethub/migrate_config_envfile.go
@@ -39,7 +39,7 @@ func (cmd *MigrateConfigEnvfileCommand) Run() error {
 		return err
 	}
 
-	output, replaceCount, err := migrateTemplateTags(string(inFileContents), refMapping, "{{ %s }}")
+	output, replaceCount, err := migrateTemplateTags(string(inFileContents), refMapping, "%s")
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


internals/secrethub/migrate_config_templates.go
```diff
@@ -48,24 +47,25 @@ func (cmd *MigrateConfigTemplatesCommand) Run() error {
...
 		return nil, ErrReadFile(inFile, err)
 	}
 
 	var hits, misses []string
 	output := regexpSecretTemplateTags.ReplaceAllStringFunc(string(raw), func(templateTag string) string {
-		path := regexpSecretTemplatePath.FindString(templateTag)
-		if path == "" {
+		path := regexpSecretTemplateTags.FindStringSubmatch(templateTag)
+		fmt.Println(path)
+		if path[1] == "" {
```
SimonBarendse, line 60: Can this panic? Should we check the length of `path` variable first?

SimonBarendse (commented):
:+1: Thanks for adding all these tests as well :raised_hands:

internals/secrethub/migrate_config_templates.go
edif2008: I tested it out and it does not look like it would panic.
I adjusted it, though, for sanity check.
Thanks for pointing it out.

florisvdg: Won't panic, but this piece of code could be made a bit more self-documenting. Instead of using the static 1 index everywhere, you could do something like was done here:
https://github.com/secrethub/secrethub-cli/blob/d9dbccfffb571127f28a3ef6a2c7c0650be05e69/internals/secrethub/migrate_config_references.go#L42-L47

florisvdg: Also, there's still a print line.

edif2008: The printline was removed with the previous commit: [Prevent regex from panicking](https://github.com/secrethub/secrethub-cli/pull/390/commits/5ba8b592e2ed461931180763efe4abd4c404c1de)

edif2008: Also, nice suggestion. I adjusted the code to be more readable

SimonBarendse (approved):

internals/secrethub/migrate_config_envfile.go
```diff
@@ -39,7 +39,7 @@ func (cmd *MigrateConfigEnvfileCommand) Run() error {
 		return err
 	}
 
-	output, replaceCount, err := migrateTemplateTags(string(inFileContents), refMapping, "%s")
+	output, replaceCount, err := migrateTemplateTags(string(inFileContents), refMapping, "{{ %s }}")
```
florisvdg, line 42: With this changed, I'm guessing that the last arg can be removed now from the `migrateTemplateTags` function signature entirely. Could you check?

florisvdg: Actually, never mind my comment: this change should be reverted, because we don't want curly braces to end up in the `.env` file.

florisvdg (approved):

PR #396, opened 2022-02-09
Add support for 1Password CLI 2 in the migration

edif2008:
Users who have the 1Password CLI 2 installed can now migrate from SecretHub to 1Password.

Note: When 1Password CLI 2 is used, it no longer adds the sign-in address in the `1password-migration-plan.yml` file and it will not check whether the address coincides with the one the user is singed in.

### DIFF
--- a/internals/onepassword/cli_v1.go
+++ b/internals/onepassword/cli_v1.go
@@ -1,0 +1,118 @@
+package onepassword
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+)
+
+type OPV1CLI struct {
+	version string
+	isV2    bool
+}
+
+func (op *OPV1CLI) IsV2() bool {
+	return op.isV2
+}
+
+func (op *OPV1CLI) CreateVault(name string) error {
+	_, err := execOP("create", "vault", name)
+	if err != nil {
+		return fmt.Errorf("could not create vault '%s': %s", name, err)
+	}
+	return nil
+}
+
+func (op *OPV1CLI) CreateItem(vault string, template *ItemTemplate, title string) error {
+	jsonTemplate, err := json.Marshal(template)
+	if err != nil {
+		return err
+	}
+
+	encodedTemplate := base64.RawURLEncoding.EncodeToString(jsonTemplate)
+
+	_, err = execOP("create", "item", "apicredential", "--vault="+vault, encodedTemplate, "title="+title)
+	return err
+}
+
+func (op *OPV1CLI) SetField(vault, item, field, value string) error {
+	_, err := execOP("edit", "item", item, fmt.Sprintf(`%s=%s`, field, value), "--vault="+vault)
+	if err != nil {
+		return fmt.Errorf("could not set field '%s'.'%s'.'%s'", vault, item, field)
+	}
+	return nil
+}
+
+// GetFields returns a title-to-value map of the fields from the first section of the given 1Password item.
+// The rest of the fields are ignored as the migration tool only stores information in the first
+// section of each item.
+func (op *OPV1CLI) GetFields(vault, item string) (map[string]string, error) {
+	opItem := struct {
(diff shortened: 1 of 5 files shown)


internals/onepassword/cli_v1.go
```diff
@@ -0,0 +1,118 @@
...
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+)
+
+type OPV1CLI struct {
+	version string
+	isV2    bool
```
hculea, line 11: I think this attribute is redundant. Instead, we should have something like 
```go
type OPV1CLI struct {
	version string
}

func (op *OPV1CLI) IsV2() bool {
	return false
}
```

internals/onepassword/cli_v2.go
```diff
@@ -0,0 +1,137 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+)
+
+type OPV2CLI struct {
+	version string
+	isV2    bool
```
hculea, line 12: same as above:
```go
type OPV2CLI struct {
	version string
}

func (op *OPV2CLI) IsV2() bool {
	return true
}
```

internals/onepassword/onepassword.go
```diff
@@ -14,55 +13,34 @@ import (
...
-	if err != nil {
-		return nil, fmt.Errorf("unexpected format of 1Password item in `op get item` command output: %s", err)
-	}
-
-	fields := make(map[string]string, len(opItem.Details.Sections[0].Fields))
-	for _, field := range opItem.Details.Sections[0].Fields {
-		fields[field.Title] = field.Value
-	}
-	return fields, nil
+	return &OPV1CLI{
```
hculea, line 40: Nit: I think we should check whether the version starts with "1." before defaulting to v1. It might be that there is something wrong with the `op` installation, and `--version` throws an error.

internals/onepassword/cli_v2.go
```diff
@@ -0,0 +1,137 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+)
+
+type OPV2CLI struct {
+	version string
```
hculea, line 11: Same comment as above:
" Do we actually ever use the version, or are we just passing it around? If not, I think making the struct, `OPV1CLI` or `OPV2CLI`, provides enough information, and we should get rid of the `version` attribute. "

hculea:
Looks good, I've left a few suggestions. 👍

internals/onepassword/cli_v1.go
```diff
@@ -0,0 +1,118 @@
+package onepassword
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+)
+
+type OPV1CLI struct {
+	version string
```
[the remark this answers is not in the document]
edif2008: I added the version because I was worried that some small changes were made to the command signature in different versions of v1. By having the version, I could easily change the signature based on the version that brought the change. Fortunately, this is not the case.

Therefore, I can easily remove the version without worrying that some version brings a breaking change in the command signature. If this will happen, we can bring the `version` back and go from there.

internals/onepassword/onepassword.go
edif2008: We can add an extra verifier for v1 for sanity check.

As for your concern, if the command throws an error (even one thrown by the CLI, e.g. invalid flag) the error will be thrown before returning the CLI client. Check lines 27-30 and `execOP` for more details.

internals/onepassword/cli_v2.go
edif2008: Indeed it can be simplified. Nice catch 😄

hculea:
Code looks good to me 👍 I'll do some functional testing later today too.

hculea (approved):
Did some functional testing too, everything seems to work well. 👍

Issue #399, opened 2022-09-13
OP_SESSION environment variable not found


G3z:
I am unable to perform the migration

```bash
$ eval $(op signin)
$ secrethub migrate plan ###
Encountered an error: OP_SESSION environment variable not found, run `eval $(op signin)` to set one
```
`OP_SESSION` variable is always empty

PR #400, opened 2022-11-08
Fix item creation with 1Password CLI 2 when migrating secrets from SecretHub

edif2008:
The bug came from the fact that 1Password CLI 2 no longer supports old item templates as input. Therefore, we now have to use the new item JSON template to make it work.

Files that fix the bug and actually require a review:
- `internals/onepassword/onepassword.go`
- `internals/onepassword/cli_v1.go`
- `internals/onepassword/cli_v2.go`
- `internals/secrethub/migrate.go`

Summary of other changes:
- Change `build` tags to newer version (i.e. from `+build` to `go:build`)
- Replace deprecated functions from `ioutil`
- Update a go dependency which enables building the CLI on arm64
- Update Go version used by the package and the pipeline to 1.18 for the following reasons:
  - Go 1.18 is safe for changing `build` tags (i.e. no change of errors for not having both tags)
  - The updated Go dependency requires Go 1.17 or higher
- Update go releaser image used in pipeline. This involved making a config change in the `.goreleaser.yaml` to be compliant to the new syntax. Check [this documentation](https://goreleaser.com/deprecations/?h=brews#brewsgithub) for understanding the change 
- Address other lint errors

### DIFF
--- a/internals/secrethub/fakes/timeformatter.go
+++ b/internals/secrethub/fakes/timeformatter.go
@@ -1,4 +1,4 @@
-// +build !production
+//go:build !production
 
 package fakes
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,13 +30,13 @@ jobs:
       arch:
         type: string
     docker:
-      - image: circleci/golang:1.18
+      - image: cimg/go:1.18
     steps:
       - checkout
       - run: GOOS=<< parameters.os >> GOARCH=<< parameters.arch >> go build ./cmd/secrethub
   test:
     docker:
-      - image: circleci/golang:1.18
+      - image: cimg/go:1.18
     steps:
       - checkout
       - restore_cache:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,13 +30,13 @@ jobs:
       arch:
         type: string
     docker:
-      - image: circleci/golang:1.17
+      - image: circleci/golang:1.18
     steps:
       - checkout
       - run: GOOS=<< parameters.os >> GOARCH=<< parameters.arch >> go build ./cmd/secrethub
   test:
     docker:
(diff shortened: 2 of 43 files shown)


internals/secrethub/fakes/timeformatter.go
```diff
@@ -1,4 +1,4 @@
-// +build !production
+//go:build !production
```
hculea, line 1: Please confirm that we don't need both of these build tags. We had an issue a while back on the 1P CLI for not including both.

hculea:
Had a look over the indicated files, change looks good to me. 

Didn't get to try it out functionally, I can fit this in next morning. If urgent, feel free to merge as is, change looks small enough (aside from the huge linter diff)

hculea (approved):

internals/secrethub/fakes/timeformatter.go
edif2008: Looking at the [tag transition](https://stackoverflow.com/a/68361186), it feels like the safest way is to upgrade to Go 1.18 and run `go fix`. This should address the concern that you have. What do you think?

hculea:
Tested together with Eddy, we identified two issues to be addressed at a later point:
* `secrethub migrate plan` enforces it finds an `OP_SESSION_*` envvar. This breaks compatibility with biometric unlock.
* `secrethub migrate apply` disregards the `sign-in address` yaml field of the generated plan.
These both can be addressed in separate issues, if need be. My approval still stands after functional testing.

jpcoenen:
Code looks good to me :+1: Thanks for fixing this, @edif2008 🙌

jpcoenen (approved):

SimonBarendse:
Does this fix https://github.com/secrethub/secrethub-cli/issues/399 ?

